PR #8, opened 2017-07-18
Editing versions, optimistic actions

Birkbjo:


app/src/components/dialog/DialogBase.js
```diff
@@ -41,8 +51,8 @@ class DialogBase extends Component {
             title={title}
             actions={actions}
             modal={false}
-            contentStyle={contentStyle || {}}
-            bodyStyle={bodyStyle || {}}
+            contentStyle={{...styles.contentStyle, ...contentStyle} || styles.contentStyle}
```
Markionium: Spreading the styles.contentStyle has the same effect as the || when contentStyle is not defined. The `|| styles.contentStyle` will never be executed. Currently this works because babel translates it to `Object.extend` which is why it will work when contentStyle is undefined. (If/when it where to use a Iterator, like the array one does, you'll need to set a defaultProp to not get an error. 

Same for the next line.

app/src/components/form/ReduxFormUtils.jsx
```diff
@@ -116,3 +116,17 @@ export const validateUploadField = (supportedExtensions = [], required) => (
 
 export const validateZipFile = validateUploadField([".zip"], true);
 export const validateImageFile = validateUploadField([".png", ".jpg", ".jpeg"]);
+
+
+const urlRegex = /^(?:(?:https?):\/\/)(?:\S+(?::\S*)?@)?(?:(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)(?:\.(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)*(?:\.(?:[a-z\u00a1-\uffff]{2,}))\.?)(?::\d{2,5})?(?:[/?#]\S*)?$/i
```
Markionium: Dat regex 😃

app/src/components/utils/Snackbar.jsx
```diff
@@ -1,32 +1,45 @@
...
 import { emptySnackbar } from '../../actions/actionCreators';
 export class Snackbar extends Component {
 
     render() {
-        const message = this.props.message;
+        const { message, retryAction } = this.props.snackbar;
+        const retryProps = {
+            action: 'Retry',
+            onActionTouchTap: retryAction
+        }
```
Markionium: Missing a semicolon. ;) Not really critical but i've seen it in a few places now. We could consider adding a linter to make sure the style is consistent.

PR #17, opened 2019-03-22
[DHIS2-5499] Add database tables to support media

erikarenhill:
With this PR there is also support when creating an app to attach an icon that gets saved to db, and a getMedia handler that can send out the file with correct content-type/mime


packages/server/migrations/20190320113929_appversionmedia.js
```diff
@@ -0,0 +1,62 @@
+
+exports.up = async (knex) => {
+
+
```
amcgee: I assume these extra line breaks at the top of functions are intentional?

erikarenhill: its an es-rule from hapijs, but we should unify this with our other styling. I can add that as a new separate Jira task.

Storing the mime type is because we currently save the file to AWS s3 and wanted this to be a technology neutral solution, e.g. if we build support for storing the files in a database (or something else) we could just push the raw bytes through and add the previously stored mime to the http header in the response without having to check the actual data. Thats my thought.

> unless we're supporting self-hosted external appstores too

this new app store is supposed to be able to run self-hosted for anyone that wants to brand it, or contain their own sets of apps for a specific instance.

amcgee:
Ok, thanks for the clarifications!

PR #18, opened 2019-03-26
[DHIS2-5931] Adding support for auth0 authorization headers

erikarenhill:
Also contains a bugfix in getMedia to return 404 if the AWS S3 object is missing.


varl:
Where is the `@utils` binding added?

erikarenhill:
It's a local alias (_moduleAlias) defined in package.json

PR #19, opened 2019-04-04
Create app with real users, add m2m api for CI integration, add links between users and organisation

erikarenhill:
This is a combined PR for the following (since they depend on each other somewhat)
DHIS2-6400
DHIS2-6044
DHIS2-5940


packages/server/seeds/01_organisation-developers.js
```diff
@@ -12,36 +12,46 @@ exports.seed = async (knex) => {
...
+        { id: 2, uuid: uuid(), name: 'World Health Organization', slug: slugify('World Health Organization', { lower: true }), created_by_user_id: 1 },
+    ])
+
+
     console.log('Inserting user-organisations #01')
     //user-organisations
     await knex('user_organisation').insert([
-        { organisation_id: 1, user_id: 2 },
-        { organisation_id: 2, user_id: 1 },
+        { organisation_id: 1, user_id: 3 }, //viktor -> dhis2
```
varl: i'm finally famous enough to have made it into a test

erikarenhill: yes, the answer to life is not 42, it's to be included into data for unit tests.

Mohammer5:
I haven't checked anything yet, just went through your commit message history and there's potential for improvement.

> feat: Store logged in users in database if they dont exist

`Store` should be lower case

> chore: create/delete organisation + tests

This is not descriptive enough, I don't know what's meant by this. Right now I'm interpreting this as someone create and delete orgs, but not sure atm

> feat: create users

DId you create users? Did you add a create user functionality?

> chore: rename file + fixed auth in getSingleApp

This is def. not `chore`. It could be either `feat` or `fix`

> chore: added try-catch around database call

Should be present, but is past

> chore: add name from auth profile

Could be `chore: add name from auth profile to AuthService`. I'm sure you can find better wording.

### Conclusion

I think we should add the commit message pre commit hook to this app. Also please try to write a bit more informative commit messages as they're a very good entry point to read the specific changes of a PR, ideally they read like a story of what you did before I dive into the code changes.

PR #20, opened 2019-04-05
[DHIS2-6395] Adds fallback to local file storage if AWS S3 is not configured

erikarenhill:


packages/server/src/utils/LocalFileSystemHandler.js
```diff
@@ -0,0 +1,124 @@
...
+const path = require('path')
+const rimraf = require('rimraf');
+
+/**
+ * Store files in local file system (app-store/packages/server/upload)
+ */
+module.exports = class LocalFileSystemHandler {
+
+    constructor() {
+        this._uploadDirectory = __dirname + '/../../upload' //app-store/packages/server/upload
```
Mohammer5: This can be a constant and doesn't need to be a class instance member.

packages/server/src/utils/LocalFileSystemHandler.js
```diff
@@ -0,0 +1,124 @@
...
+    saveFile(directoryPath, filename, buffer) {
+        return new Promise((resolve, reject) => {
+            try {
+                this.verifyPath(directoryPath)
+                this.createDirectoryIfNotExists(path.join(this._uploadDirectory, directoryPath))
+            } catch ( err ) {
+                reject(err)
+            }
+            
+            fs.writeFile(path.join(this._uploadDirectory, directoryPath, filename), buffer, (err) => {
```
Mohammer5: Afaik this code will still be executed in case of promise rejection, not 100% sure though.
If this is intended, nevermind :)

packages/server/src/utils/LocalFileSystemHandler.js
```diff
@@ -0,0 +1,124 @@
...
+            } catch ( err ) {
+                reject(err)
+            }
+
+            fs.readFile(path.join(this._uploadDirectory, directoryPath, filename), (err, data) => {
+                if ( err ) {
+                    console.log(err)
+                    reject(err)
+                } else {
+                    resolve({ Body:data })
```
Mohammer5: You can just change the parameter name to `Body` and then do `resolve({ Body })`, besides that, there's a spacing missing, making this assignment look a bit weird

packages/server/src/utils/LocalFileSystemHandler.js
```diff
@@ -0,0 +1,124 @@
...
+     */
+    deleteFile(directoryPath, filename) {
+        return new Promise((resolve, reject) => {
+            try {
+                this.verifyPath(directoryPath)
+            } catch ( err ) {
+                reject(err)
+            }
+
+            fs.unlink(path.join(directoryPath, filename), (err) => {
```
Mohammer5: Same as above, might still get executed and trigger non-blocking errors under the hood

packages/server/src/utils/LocalFileSystemHandler.js
```diff
@@ -0,0 +1,124 @@
...
+    verifyPath(directoryPath) {
+        if ( directoryPath.indexOf('..') !== -1 ) {
+            throw new Error('Relative paths not allowed')
+        }
+
+        if ( directoryPath.indexOf('~') === 0 ) {
+            throw new Error('Tilde not allowed in directoryPath')   //avoid reading from current user directory
+        }
+
+        if ( directoryPath.indexOf('/') === 0 ) {
```
Mohammer5: I'm confused, what kind of path can it be if it can't be relative and can't be absolute?

packages/server/src/utils/createBackingStorageInstance.js
```diff
@@ -0,0 +1,26 @@
+const AWSFileHandler = require('./AWSFileHandler')
+const LocalFileSystemHandler = require('./LocalFileSystemHandler')
+
+/**
+ * Factory function returning an instance of the backing file storage to be used
+ */
+const createBackingStorageInstance = () => {
+
+    let storage = null
```
Mohammer5: Nothing major, but there's no need for this assignment, you could just return the newly created instances

packages/server/src/utils/LocalFileSystemHandler.js
erikarenhill: it should be relative but not able to escape up to it's parent outside of the allowed directory using '..', that's what I meant with the description at least, or if someone somehow manages to pass down an absolute path it should not be able to save/read files to other places on disk

packages/server/src/utils/LocalFileSystemHandler.js
erikarenhill: would be possible, but this is essentially just to return on the same format as the AWS api with the data in a Body property, I'd like to keep the name 'data' for the callback parameter as I believe that's cleaner and that you also see the assignment in the resolve call. Agree about the whitespace though

PR #21, opened 2019-04-08
[DHIS2-6553] optional authentication

erikarenhill:
This makes it possible to run the backend without auth0 using unprotected endpoints making dev-life easier. It currently makes the app-store backend possible to run configuration-less (provided you have seeded the database)


erikarenhill:
No particular reason. Ill change it separately since there are some other
vars lowercase too.

PR #22, opened 2019-04-09
docs: fixed + improved jsdocs for all db functions

erikarenhill:


varl:
>But as previously stated, you can read it in the function declaration and stating it again in the JSDocs makes it redundant? Lets move the JSDocs comments to the PR, belongs more there than here..

You're right. It's redundant to use the `@async` tag in the doc string. If you give your function a name, JSDoc is also able to infer that it is a `@function` and what the `@name` is, so those tags would also be redundant if we used the old version of the code.

In the PR version of the code, I cannot read the function name in the function declaration, it is still 10 or so lines above the function inside of the doc string. And if I'm up there trying to read the name, might as well have the information that it's an async function in the same place so my eyeballs do not have to jump up and down a handful of lines to hunt for information which could be served in the same place.

I find the name of a function critical to understanding the intent of a function so it is quite important to me to be available in the surrounding context of the function body.

Otherwise I'd just name all my functions `foo` or `a` because the user of the function should give it another name when s/he imports it.

---

I still prefer the initial version:

```js
const addAppVersionMediaAsync = async (params, knex, transaction) => {...}
```

... but without the `Async` suffix:

```js
const addAppVersionMedia = async (params, knex, transaction) => {...}
```

varl:
Every single async file is inside of `src/data`. What if we just rename `data` to `async` and do this:

```
async/
├── addAppVersionMedia.js
├── addAppVersionToChannel.js
├── addUserToOrganisation.js
├── createApp.js
├── createAppStatus.js
├── createAppVersion.js
├── createLocalizedAppVersion.js
├── createOrganisation.js
├── createUser.js
├── deleteApp.js
├── deleteOrganisation.js
├── getAllAppsByLanguage.js
├── getAppsByStatusAndLanguage.js
├── getAppsByUuidAndStatus.js
├── getAppsByUuid.js
├── getOrganisationByUuid.js
├── getOrganisationsByName.js
├── getUserByEmail.js
└── index.js
```

You could even change the `require('@data/getUserByEmail')` to `require('@async/getUserByEmail')` for very concise indications that what you want to import is an async function.

Or am I missing the point and it's actually when calling the function it is important to have the `Async` suffix? E.g. `await getUserByEmailAsync()`?

Mohammer5:
I'm wondering if this is ultimately defying the purpose. You're opting for better readability of the code for one specific problem (not knowing if a function is sync or async), but it's generally much harder to read function definitions.

The changes you've introduced in this PR create the following three problems to me:
1. Functions don't have names anymore (interpreting a call stack when debugging is a lot harder)
1. The exports are not named anymore `=>` When refactoring a function (and it's name to match the changes), you'll have to edit every import statement manually (of course this can be done with search and replace to some degree) but you might end up with variable names that don't match the functionality anymore

And these problems apply to the whole codebase, not just async functions, so there'll be many more instances where you'll encounter this then wondering whether a function is async or not.
I think it's totally valid to suffix functions with `Async` when working in a node/server environment as there are generally many things that can be done both synchronously and asynchronously.

@varl 's idea of having an `async` namespace is interesting, but I think it wouldn't solve the problem as what you want is to not have to think about it anymore when reading the code => reducing the cognitive overhead

erikarenhill:
> Or am I missing the point and it's actually when calling the function it is important to have the Async suffix? E.g. await getUserByEmailAsync()?

That is actually the point, but its probably a legacy from me also having worked with .NET-backends the past years where that's standard naming-convention.

I'll adhere to your suggestions and move the function name into the export

PR #67, opened 2019-09-10
feat: endpoint for creating new versions to existing app

erikarenhill:
* fixes DHIS2-7509

Compatible/legacy v1 endpoint that works with existing frontend client, also tested from commandline with curl post. Still needs some improved error handling in the request handler for example if you try to upload a version with an existing version-number.


erikarenhill:
fixes DHIS2-7509

PR #82, opened 2019-09-24
webpack config tweaks + add cypress as dev deps

erikarenhill:
fixes so 'npm run dev' works again after FE compilation was moved to internal startup with the node backend.

Also adds some upgrades to package-lock.json and add cypress as dev-deps for the upcoming tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "scripts": {
         "build": "webpack -p --config ./webpack.config.js",
         "start": "NODE_ENV=production node ./src/main.js",
-        "dev": "DEBUG=knex:tx NODE_ENV=development nodemon --inspect src/main.js",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -8,12 +8,13 @@
     "scripts": {
         "build": "webpack -p --config ./webpack.config.js",
         "start": "NODE_ENV=production node ./src/main.js",
-        "dev": "DEBUG=knex:tx NODE_ENV=development nodemon --inspect src/main.js",
+        "dev": "DEBUG=knex:tx NODE_ENV=development DEBUG='appstore*' nodemon --watch ./src --watch ./client/ --inspect src/main.js",
```
varl: You could add the `appstore*` to the first `DEBUG` variable so you don't overwrite the `knex:tx`:

```
DEBUG=knex:tx,appstore*
```

PR #100, opened 2019-10-03
fix: add rollback to createApp-handler

Birkbjo:
If an error occured when creating an app, the next request to create it was hanging. This was due to missing rollback. However, `0.17` version of knex did not support this way of using rollback, as the rollback() call was causing an unhandled rejection - which I tried to catch but to no avail - it was still crashing hapi. After hours of debugging and banging my head against the wall, I discovered this [issue with knex](https://github.com/tgriesser/knex/issues/3246). So I updated knex which seems to fix the problem.

**Do not merge before** #108


src/routes/v1/apps/handlers/createApp.js
```diff
@@ -242,6 +242,7 @@ module.exports = {
         } catch (err) {
             debug('ROLLING BACK TRANSACTION')
             debug(err)
+            trx.rollback()
```
varl: If we need to manually call `trx.rollback()` here, then we should need to do so in every handler as this pattern is widely used.

From https://knexjs.org/#Transactions:

> Throwing an error directly from the transaction handler function automatically rolls back the transaction, same as returning a rejected promise.
>
> Notice that if a promise is not returned within the handler, it is up to you to ensure trx.commit, or trx.rollback are called, otherwise the transaction connection will hang.
>
> Calling trx.rollback will return a rejected Promise. If you don't pass any argument to trx.rollback, a generic Error object will be created and passed in to ensure the Promise always rejects with something.

Birkbjo: Yup. Or we need to change to the transaction-handler syntax, which automatically rollsback if something is thrown inside the handler. 
It might automatically roll-back if part of the transaction fails. The reason for this rollback was due to validation of the appVersion, ie. an error thrown from our data-handler.

varl: @erikarenhill thoughts on this? I did use explicit `commit`/`rollback` at one point but ran into problems as we need the transaction to last for the life-cycle of an entire query. Not sure why but I seem to have thought that going for the "throw an error from the handler" option would work:

https://github.com/dhis2/app-store/commit/939f19f2ac6b45197f60a9de32a8712de9938236

varl: > Either way, the safest choice is to always have a commit and rollback for every transaction.

Yeah, what we do want is to use one transaction for the entire `createApp` query, which might call multiple other mutations for the database, so that if _anything_ fails, _all of it_ fails. We need to make sure that works.

Birkbjo: > Not sure why but I seem to have thought that going for the "throw an error from the handler" option would work: 939f19f
Well that's how it works if you use the handler-syntax, so it's not that weird.

> Yeah, what we do want is to use one transaction for the entire createApp query, which might call multiple other mutations for the database, so that if anything fails, all of it fails. We need to make sure that works.
This is how it currently is for `createApp` atleast. You have one transaction that is passed to every "mutation"-query. As far as I've tested it seems to work.

Birkbjo:
Regarding update of hapi, I noticed we use the wrong, deprecated version. We should use `@hapi/hapi` and `@hapi/joi`. I will update these in another PR, as it's some work due to some API changes in Joi...

src/routes/v1/apps/handlers/createApp.js
Birkbjo: Another thing regarding this, I feel that the handler-methods are way too big atm. Shouldn't we try to avoid too much business-logic in the handlers directly? Just a small abstraction where we just pass the database, so that handles all the transaction stuff? For instance, we could just move most of the code for ´createApp´-handler to the ´data/createApp.js´ file. That takes care of creation of all dependent entities. I'm no backend-guy, but right now we have one anonymous function with over 200 lines.

Doesn't necessarily need to refactor, but something we could look into for v2-api?

varl: >Shouldn't we try to avoid too much business-logic in the handlers directly?

Yeah, that would be nice to keep separate from the handlers.

erikarenhill: > @erikarenhill thoughts on this? I did use explicit `commit`/`rollback` at one point but ran into problems as we need the transaction to last for the life-cycle of an entire query. Not sure why but I seem to have thought that going for the "throw an error from the handler" option would work: [939f19f](https://github.com/dhis2/app-store/commit/939f19f2ac6b45197f60a9de32a8712de9938236)

yes when you mention it I do remember this discussion back then and wanting to keep a "request"- transaction rather than having to remember to do this manually. If we can get the auto/rollback on throw feature I'd prefer that, however it would also require more knowledge about how knex works internally for every developer that want to work with the app store (e.g. just reading the code and now knowing or as me forgot about this feature would think it is missing)

erikarenhill: > > Shouldn't we try to avoid too much business-logic in the handlers directly?
> 
> Yeah, that would be nice to keep separate from the handlers.

Totally agree. The new app store backend started from scratch and iterated forward in small steps where handlers grew as a proof of concept to a state where it is today. Now is the time when we should break out the logic from handlers and refactor that.

erikarenhill:
I think this PR can be deleted with regards to this: https://github.com/dhis2/app-store/pull/131

PR #101, opened 2019-10-03
chore: use webpack-dev-server for dev

Birkbjo:
Adds webpack-dev-server.
Waiting for the backend to compile the frontend for every change was annoying. I feel that running webpack-dev-server and running the backend with `USE_PREBUILT_APP` is a much smoother developer experience, as you get hot-reloading. With this, I don't see the need for the backend to restart for every client-change, so I removed that.


package.json
```diff
@@ -8,7 +8,8 @@
     "scripts": {
         "build": "webpack -p --config ./webpack.config.js",
         "start": "NODE_ENV=production node ./src/main.js",
-        "dev": "DEBUG=knex:tx,appstore* NODE_ENV=development nodemon --watch ./src --watch ./client/ --inspect src/main.js",
+        "dev": "DEBUG=knex:tx,appstore* NODE_ENV=development nodemon --watch ./src --inspect src/main.js",
+        "fe:dev": "NODE_ENV=development webpack-dev-server",
```
varl: We can use `concurrently` to do this:

```
dev               concurrently "npm:dev:*"
dev:server        USE_PREBUILT_APP=1 DEBUG=knex:tx,appstore* NODE_ENV=development nodemon --watch ./src --inspect src/main.js
dev:client        NODE_ENV=development webpack-dev-server
```

If you want to run everything in production, just use `npm start`.

If you only care about the server, then use `npm run dev:server`

If you only care about the client, you still _need_ the server unless you are testing failure modes: `npm run dev`

package.json
```diff
@@ -8,7 +8,9 @@
     "scripts": {
         "build": "webpack -p --config ./webpack.config.js",
         "start": "NODE_ENV=production node ./src/main.js",
-        "dev": "DEBUG=knex:tx,appstore* NODE_ENV=development nodemon --watch ./src --watch ./client/ --inspect src/main.js",
+        "dev": "concurrently npm:dev:*",
```
varl: Oh, add the `USE_PREBUILT_APP=1` to this command and it should work.

PR #108, opened 2019-10-08
fix: update hapi with dependencies

Birkbjo:
We used the wrong npm packages for hapi. Moved to namespace `@hapi/` now. 
Most of the changes are due to Joi API-changes. See https://github.com/hapijs/joi/issues/2037 for changelog. 

TL;DR breaking changes that I've updated;
- `Joi.validate()` removed, need to use `schema.validate(obj)`
- `validation.error` is undefined and not `null` if validation passes.
- Validation messages changed. Only one test relied on this though.
- Import paths
- `joi.allow()`, `joi.valid()` and friends does not support arrays anymore (variable args instead). Stupid change imo. I just spread the arrays to comply.


src/data/addAppVersionMedia.js
```diff
@@ -50,7 +50,7 @@ const paramSchema = joi
 const addAppVersionMedia = async (params, knex, transaction) => {
     const validation = paramSchema.validate(params)
 
-    if (validation.error !== null) {
+    if (validation.error !== undefined) {
```
varl: I assume there's a reason we simply do not `if (validation.error)` ?

Birkbjo: I just changed it to this due to the ease of find and replacing it. I'm sure we could do that as well.

package.json
```diff
@@ -22,6 +22,8 @@
         "@babel/core": "^7.6.2",
         "@babel/preset-env": "^7.4.5",
         "@babel/preset-react": "^7.0.0",
+        "@hapi/code": "^6.0.0",
```
erikarenhill: this should be a dev-dependency?

package.json
```diff
@@ -82,12 +82,12 @@
     },
     "devDependencies": {
         "@dhis2/cli-style": "^4.1.1",
-        "code": "^5.2.4",
+        "@hapi/joi": "^16.1.7",
```
erikarenhill: I think joi is supposed to go as a normal dep and not devDependency

Birkbjo: Good catch! Apparently my I confused code and joi when moving them.

PR #109, opened 2019-10-08
feat: don't wait for app to compile before api-init

Birkbjo:
Small quality of life change. Start API-server immediatly instead of waiting for app-compilation.

Worst case is that you get 404-error if the compilation is not finished yet.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -16,12 +16,12 @@ process.on('unhandledRejection', err => {
     process.exit(1)
 })
 
-compile()
-    .catch(err => {
-        debug('The web app failed to compile.\n', err)
-        process.exit(1)
-    })
-    .then(() => migrate(knex))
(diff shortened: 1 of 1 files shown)


varl:
The reason the API waits for the App to build is because if the App build fails, we do not want EBS to switch over traffic to the node in the blue/green deployment model we have set up on AWS.

With this change, if the API boots up the EBS healthcheck will see 200 OK for end-points and direct traffic to it. If the App then fails to compile, then we have a broken deployment in production.

To merge this change we have to change the healthchecks for EBS as well.

Birkbjo:
Makes sense. Just something I thought would be nice if there were no bad repercussions.

varl:
@Birkbjo sure, we can still do this, we just need to do it in a different order and change how the healthchecks in EBS work.

PR #111, opened 2019-10-09
feat: support to add and edit release channel for an app version

erikarenhill:


client/src/components/form/UploadAppFormStepper.jsx
```diff
@@ -66,6 +66,13 @@ const validate = values => {
     ) {
         errors.version.minVer = 'Cannot be higher than maximum version'
         errors.version.maxVer = 'Cannot be lower than minimum version'
+    } else if (values.version) {
+        if (!values.version.minVer) {
```
Birkbjo: Any reason not to include these fields in `requiredFields` above?

client/src/components/form/UploadAppFormStepper.jsx
```diff
@@ -139,19 +150,27 @@ const AppVersionSection = props => {
...
+                label="Maximum DHIS version *"
             >
                 {DHISVersionItems}
             </Field>
             <br />
+            <Field
+                name="channel"
+                component={formUtils.renderSelectField}
+                label="Release channel *"
+            >
```
Birkbjo: Should we have a default value?

test/data/index.js
```diff
@@ -295,4 +295,50 @@ describe('@data::updateAppVersion', () => {
...
+        )
+
+        await transaction.commit()
+        ;[app] = (await getAppsByUuid(mockAppUuid, 'en', db)).filter(
+            app => app.version_uuid === appVersionUuidToUpdate
+        )
+        expect(app.channel_name).to.equal('Development')
+
+        //Try to change to a channel that doesn't exist
+        transaction = await db.transaction()
```
Birkbjo: This should be it's own test.

package.json
```diff
@@ -43,13 +44,12 @@
...
         "html-webpack-plugin": "^3.2.0",
         "inert": "^5.1.2",
-        "joi": "^14.3.1",
         "jwt-decode": "^2.1.0",
         "knex": "^0.17.3",
         "lodash": "^4.17.15",
         "material-components-web": "^0.16.0",
         "material-design-icons": "^3.0.1",
-        "material-ui": "^0.18.7",
+        "material-ui": "^0.20.2",
```
Birkbjo: This is risky. I know that there was a dropdown bug when upgrading, even though that was a year ago or so.

Birkbjo: Just tested this. If you click a dropdown it will flicker to the top left.

client/src/components/form/UploadAppFormStepper.jsx
erikarenhill: Not really, it was already in a required-array but looking again I see they were originally in the required array for varChar fields which they arn't

client/src/components/form/UploadAppFormStepper.jsx
erikarenhill: Personally I think it's better to require an active choice to not publish an app to the wrong channel, the same way we require selecting max/min version.

package.json
erikarenhill: Good memory, I'll revert

client/src/components/form/UploadAppFormStepper.jsx
varl: I agree that an active choice would be OK.

More importantly, does this affect the App Management application when selecting apps for now or can it use the old way?

erikarenhill: The v1 "apps" api for fetching apps still just returns the stable apps. Currently no v2 api endpoint for fetching another release channel is implemented. Will merge this first and then  tweak this behaviour and add v2 endpoint for selecting a channel

PR #114, opened 2019-10-10
fix: contentype to json

Birkbjo:
After upgrading hapi, something about how it handles `content-type` has changed, resulting in PUT-requests failing since they were sending the JSON as `text/plain`. The error was due to sending a JSON-stringified object to Joi-validation. Before it seems that hapi would be able to parse this into `request.payload` automatically, but not always? It's weird, as some endpoints (`/editAppVersions`, `/editApp`) did not need to parse the payload, but `/editImage` did? ~~I don't really understand why. Any ideas @erikarenhill ?~~

Either way I feel that it's a good thing that the error was caught, as it shouldn't really have worked before.

**EDIT**: I figured out why: `joi@16` is no longer automatically converting strings to object or arrays, see "Array and object string coercion" in the release notes: https://github.com/hapijs/joi/issues/2037.
 `/editImage` was not using joi-validation so that is why we needed to parse the input.

### DIFF
--- a/client/src/api/api.js
+++ b/client/src/api/api.js
@@ -16,6 +16,9 @@ const deleteOpts = {
 
 const updateOpts = {
     method: 'PUT',
+    headers: {
+        'content-type': 'application/json',
+    },
 }
 
 export function getAllApps() {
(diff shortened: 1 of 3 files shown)


Birkbjo:
Note for self: Should set`routes.payload.allow` to `application/json`, so by default we only allow json payloads. Multipart routes should already override this.

PR #117, opened 2019-10-14
chore(deps): remove mock-knex and sqlite3

erikarenhill:
Removed two dependencies that we don't use, mock-knex and sqlite3


erikarenhill:
I'll just delete this PR instead since we will look into mock-knex again to be able to automate running tests in our aws pipeline.

PR #122, opened 2019-10-15
feat: Release channel filter on apps endpoint

erikarenhill:
Added a query-parameter filter option on the v1 apps endpoint with default fallback to Stable channel. As this is mere a filtering option I decided to not implement this feature into a v2 endpoint.

DHIS2 instances could therefore just add ?channel=Canary to the appstore backend url to switch to Canary channel for example


erikarenhill:
@varl Hmm, I think this feature could be merged, however we need to revisit how to handle this against the app management app, since that only has the base url and could not use querystrings. Then we would need to also provide something like appstore.api.channel next to appstore.api.version config

varl:
> @varl Hmm, I think this feature could be merged, however we need to revisit how to handle this against the app management app, since that only has the base url and could not use querystrings. Then we would need to also provide something like appstore.api.channel next to appstore.api.version config

You mean that this would be configured per instance in `dhis.conf` to use e.g. canary or development builds only?

erikarenhill:
> > @varl Hmm, I think this feature could be merged, however we need to revisit how to handle this against the app management app, since that only has the base url and could not use querystrings. Then we would need to also provide something like appstore.api.channel next to appstore.api.version config
> 
> You mean that this would be configured per instance in `dhis.conf` to use e.g. canary or development builds only?

Either that, or managed by adding a filter in the app management app where you could select release channel, then we just need a v2 endpoint to fetch existing channels and their names

varl:
I think that it would be more powerful to be able to pick and choose which apps to run on stable and which to run on development. That should also allow users to use e.g. the stable and canary version side-by-side (if they want that).

erikarenhill:
> I think that it would be more powerful to be able to pick and choose which apps to run on stable and which to run on development. That should also allow users to use e.g. the stable and canary version side-by-side (if they want that).

I see the point where someone want to use one single app from canary (to try it out) but the rest of the apps from Stable, and then using the channel in dhis.conf would be a bad choice. I was just about to edit my previous comment when I saw your response. Yes this is the way to go.

erikarenhill:
Either way, this PR supports moving towards that goal and is also backwards compatible

PR #124, opened 2019-10-16
ci: install dev deps when building in pipeline

varl:


varl:
Yeah, this build in the pipeline is just to verify that it does build. The artifact is not stored and re-used by the production/stage envs.

PR #126, opened 2019-10-16
fix: add missing await to async commits

erikarenhill:
there were multiple async transaction commits that didn't wait for completion


varl:
I don't remember what it was, but when I did this in June I didn't get the expected rollback behaviour. This is how I would expect it to work though so if it's tested and confirmed to roll back correctly in the database I'm good with this.

erikarenhill:
> I don't remember what it was, but when I did this in June I didn't get the expected rollback behaviour. This is how I would expect it to work though so if it's tested and confirmed to roll back correctly in the database I'm good with this.

ah, I'll check and verify, however the tests pass successfully (which doesnt necessarily mean it works.... we could lack tests of that)

erikarenhill:
> > I don't remember what it was, but when I did this in June I didn't get the expected rollback behaviour. This is how I would expect it to work though so if it's tested and confirmed to roll back correctly in the database I'm good with this.
> 
> ah, I'll check and verify, however the tests pass successfully (which doesnt necessarily mean it works.... we could lack tests of that)

From knex docs:
`Notice that if a promise is not returned within the handler, it is up to you to ensure trx.commit, or trx.rollback are called, otherwise the transaction connection will hang.`

Bases on this I assume it will auto-commit as well if everything goes well, opposed to rollback if something fails. However if we want to do things based on if the database transaction succeeds we must await the commit (for example deleting a file on disk only if the commit is successful).

I will do a test to verify

src/routes/v1/apps/handlers/deleteApp.js
```diff
@@ -42,7 +42,7 @@ module.exports = {
         const trx = await knex.transaction()
         try {
             await deleteApp(appUuid, knex, trx)
-            trx.commit()
+            await trx.commit()
```
erikarenhill: If we dont wait for the commit it would delete the files even if the db-transaction fails

src/routes/v1/apps/handlers/deleteImage.js
```diff
@@ -49,7 +49,7 @@ module.exports = {
 
                 await deleteMedia(media_uuid, db, transaction)
 
-                transaction.commit()
+                await transaction.commit()
```
erikarenhill: If we dont wait for the commit it would delete the files even if the db-transaction fails

Birkbjo:
> > > I don't remember what it was, but when I did this in June I didn't get the expected rollback behaviour. This is how I would expect it to work though so if it's tested and confirmed to roll back correctly in the database I'm good with this.
> > 
> > 
> > ah, I'll check and verify, however the tests pass successfully (which doesnt necessarily mean it works.... we could lack tests of that)
> 
> From knex docs:
> `Notice that if a promise is not returned within the handler, it is up to you to ensure trx.commit, or trx.rollback are called, otherwise the transaction connection will hang.`
> 
> Bases on this I assume it will auto-commit as well if everything goes well, opposed to rollback if something fails. However if we want to do things based on if the database transaction succeeds we must await the commit (for example deleting a file on disk only if the commit is successful).
> 
> I will do a test to verify

I believe this is just if you use the handler-syntax, ie ```
knex.transaction(async trx => await trx.insert()...)```.

erikarenhill:
> > > > I don't remember what it was, but when I did this in June I didn't get the expected rollback behaviour. This is how I would expect it to work though so if it's tested and confirmed to roll back correctly in the database I'm good with this.
> > > 
> > > 
> > > ah, I'll check and verify, however the tests pass successfully (which doesnt necessarily mean it works.... we could lack tests of that)
> > 
> > 
> > From knex docs:
> > `Notice that if a promise is not returned within the handler, it is up to you to ensure trx.commit, or trx.rollback are called, otherwise the transaction connection will hang.`
> > Bases on this I assume it will auto-commit as well if everything goes well, opposed to rollback if something fails. However if we want to do things based on if the database transaction succeeds we must await the commit (for example deleting a file on disk only if the commit is successful).
> > I will do a test to verify
> 
> I believe this is just if you use the handler-syntax, ie ` knex.transaction(async trx => await trx.insert()...)`.

yes! This is the conclusion we came up to as well, so we shall use await and explicit commit/rollback everywhere.

PR #127, opened 2019-10-16
feat: better support for versioning api's with fallback

erikarenhill:
requesting /api/xxxx now defaults to /api/v1/xxxx through the use of new library `hapi-api-version`


erikarenhill:
> Nice.

Though... if I recall this plugin might have caused issues with hapi-swagger, which is broken now anyway so we'll have to test this again when hapi-swagger has updated to support joi 16.x

PR #133, opened 2019-10-22
fix: avoid overwrite headers using PUT and auth

erikarenhill:
The recently added headers for content-type application/json editApp / PUT request would overwrite authorization headers when used in combination, causing status 401 when trying to edit an app. This makes a check and merges the headers instead of overwriting.


client/src/api/api.js
```diff
@@ -117,9 +117,12 @@ export function deleteReview(appId, reviewId) {
...
 export function fromApi(url, auth = false, extraOpts) {
     const headers = getAuthHeaders()
-    const opts = auth
-        ? { headers, ...baseOptions, ...extraOpts }
-        : { ...baseOptions, ...extraOpts }
+    if (auth && extraOpts) {
+        extraOpts.headers = { ...headers, ...extraOpts.headers }
+    } else if (auth) {
+        extraOpts = { headers }
+    }
```
varl: Could we avoid mutating the parameters passed to the function for readability?

PR #135, opened 2019-10-24
feat: add possibility to filter apps by channel

erikarenhill:
Adds a channel filter to the main appcard-view which also merges in app type filtering into the same popover. Also displays channel as a new column in the versions table at the bottom of the the app detail view


client/src/constants/actionTypes.js
```diff
@@ -73,3 +73,7 @@ export const APP_IMAGE_SET_LOGO_SUCCESS = 'APP_IMAGE_SET_LOGO_SUCCESS'
 export const APP_IMAGE_SET_LOGO_ERROR = 'APP_IMAGE_SET_LOGO_ERROR'
 
 export const SNACKBAR_EMPTY = 'SNACKBAR_EMPTY'
+
+export const CHANNELS_LOADED = 'CHANNELS_LOADED'
+export const CHANNELS_LOAD = 'CHANNELS_LOAD'
+export const CHANNELS_ERROR = 'CHANNELS_ERROR'
```
Birkbjo: I know the app is not very consistent with action names, but I like the `SUCCESS` and `ERROR` postfix. We also use those prefixes for optimistic actions, so it would be nice follow - even though this couldn't be optimistic.

client/src/reducers/formReducer.js
```diff
@@ -93,6 +93,27 @@ const form = formReducer.plugin({
...
+            case actions.CHANNELS_LOADED: {
+                const channels = action.payload
+                const filterValues = channels.reduce((acc, c) => {
+                    acc[c.name] = c.name === 'Stable'
+                    return acc
+                }, {})
+                return {
+                    ...state,
+                    values: {
+                        ...filterValues,
```
Birkbjo: I know I wrote this, but this could just be `values: filterValues`. I was trying to merge the current values in, but that is undefined at this point. And I forgot to change it back.

client/src/constants/actionTypes.js
erikarenhill: @Birkbjo so CHANNELS_LOAD_SUCCESS instead of CHANNELS_LOADED for example, right? I agree in that case, will change it

PR #137, opened 2019-10-30
refactor: separate static/api to hapi plugins

erikarenhill:
To get better separation in the code, and also isolating the hapi server context between static and api routes this PR moves them into separate plugins to clean up a bit and hopefully work better

### DIFF
--- a/src/plugins/apiRoutes.js
+++ b/src/plugins/apiRoutes.js
@@ -1,0 +1,57 @@
+const jwt = require('hapi-auth-jwt2')
+
+const registerAuth0 = require('../security/registerAuth0.js')
+
+const routes = require('../routes/index.js')
+
+const apiRoutesPlugin = {
+    name: 'DHIS2 App Bazaar Backend',
+    register: async (server, options) => {
+        const { knex } = options
+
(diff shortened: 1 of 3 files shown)


src/plugins/apiRoutes.js
```diff
@@ -0,0 +1,57 @@
...
+                    '\x1b[41m',
+                    'Running without authentication requires to setup mapping to a user to use for requests requiring a current user id (e.g. creating apps for example). Set process.env.NO_AUTH_MAPPED_USER_ID',
+                    '\x1b[m'
+                )
+                process.exit(1)
+                return
+            }
+        }
+
+        server.realm.modifiers.route.prefix = '/api'
```
varl: Will this modify the root `server` object and therefore give us the same behaviour?

varl:
Let's try it.

src/plugins/apiRoutes.js
erikarenhill: If I understand the docs correctly the server sent into each plugin-registration will be isolated/scoped to that plugin which will resolve that previously discussed issue doing it this way.

PR #145, opened 2019-11-11
feat: add dhis2 version filter support in api

erikarenhill:
* fixes DHIS2-7829


erikarenhill:
> LGTM. I've made a small filter "plugin" that moves all specified filters into a filter object instead of having to use the query-params.
> However I don't think it will be a big change if we start using that.

I'll merge this in now and we can start using that, feel free to add a PR so I can see how it's supposed to be used!

PR #146, opened 2019-11-12
test: add tests to increase coverage

erikarenhill:
Added more tests generically to increase coverage on untested code + some fixes on bugs discovered meanwhile adding the tests.

We still need to improve coverage, but with this PR we at least pass above 80% (from 78%)


src/data/getUserByEmail.js
```diff
@@ -9,6 +9,14 @@ const debug = require('debug')('appstore:server:data:getUserByEmail')
 const getUserByEmail = async (email, knex) => {
     let user = null
 
+    if (!email) {
+        return null
+    }
+
+    if (!knex) {
+        throw new Error(`Missing parameter 'knex'`)
```
Birkbjo: We could look into something like https://hapi.dev/family/bounce/?v=1.3.2 to avoid having these everywhere. I guess the reason is due to the catch below which would silently fail due to the 'undefined' error that would otherwise occur, right?

erikarenhill: will look into bounce!

PR #154, opened 2019-11-19
feat: rename channel endpoint and new auth config

erikarenhill:
This is a bit PR since it's actually two semantic changes in one, but it just went this way with the refactoring of the new authentication registrations and strategy/schemes.

**The big change** here is that instead of reading out process.env- variables all over the place, we now build a configuration object and pass to the init (server) function. Also the authentication in the handlers is now either required, optional or false instead of calling a function requiring different auth depending on env vars. What happens instead is that the custom strategies required and optional is created differently (apiRoutes.js) to either use auth0 or allow everything. 

This way we can easier pass a configuration object from our unit tests opening up authenticated endpoints for tests without messing with auth0 calls in there.

Then there's a new feature where I've added an endpoint for updating a channels name..


erikarenhill:
@varl can you have an initial look at this before I wrap this up? I've moved process.env stuff into a config object instead and making custom auth-strategies for running fully unauthenticated instead of reading env vars all over the place to decide different strategies. This way we can pass different configs when initializing the server from tests/etc instead of relying of env vars for different setups (still needed for default configuration with rds, auth0 etc.)

varl:
Makes sense. 👍

PR #156, opened 2019-11-22
fix: throw error from apiPlugin instead of exit

Birkbjo:
Plugins should not call System.exit(). When running tests without `NO_AUTH_MAPPED_USER_ID` I did not get an error, the tests just exit as debug are not shown in tests, which was very confusing. To me it makes much more sense to throw an error here, which also crashes the app.


varl:
The reason for the `process.exit(1)` is that when the App exits with an error code, then EBS tries to restart the process. If the error causes the process to exit with `1` then this is OK.

erikarenhill:
> The reason for the process.exit(1) is that when the App exits with an error code, then EBS tries to restart the process. If the error causes the process to exit with 1 then this is OK.

Then we could simple check the env and throw an error if we're running tests

varl:
@Birkbjo this can be rebased on master and merged.

Issue #139, opened 2019-10-30
Notes about architecture and structure


Birkbjo:
I've been working with the backend the last few weeks, and in that regard I have some notes and feedback I want to discuss.

- Model concept
Currently we have no concept of database "models", they are merely validations. I don't know if we need these, but I've been fiddling with the idea of having and "internal" model, which is basically a combination of a Joi-schema and some helper methods and constants for that "model". Does not need to be a class, just a file exporting helpers for that entity. Like having a map of internal fields to database fields, and a way to parse from and to the database. These could for instance be in root of 'models/v2'.

- The 'data' concept
   I feel these cumbersome to use, as we have to parse the data anytime it's used. I feel like these can be grouped as a 'service' for each entity, and in combination with the 'models', it would be easier to use internally. With this I mean that we would never 'expose' the result of the database queries directly, they would be parsed to an internal structure, so we don't need to care about column names and collecting joined relations anytime data is fetched.

- Validation
 Currently we are validating in both routes and 'data', I belive this is overkill. It should probably be up to the route handlers to do validation through hapi. Internal wrong usage is developer error and we shouldn't use Joi for type-checking.

- Transactions
  I've been fiddling with the 'service'-idea, and I believe if we do this right we can cut down on the 'try-catching' of transactions. Knex supports using transactions in the same way you would've used `knex()`, so we can use these interchangeably. This means that we can use the transaction-handler syntax in route-handlers and compose service-methods by passing either the transaction or `knex`-object - with the benefit of automatic rollback. Rough concept:
```javascript 
// some handler
const createVersionAndAddToChannel = (trx => {
     const v = await Version.create(payload.version, trx);
     await Channel.addVersion({appVersionId: v.id, channelName: payload.channel, ...}, trx)
     return v;
})
const version = h.db.transaction(createVersionAndAddToChannel);
```

We probably still need some explicit transaction management due to the normalized structure of our database, and some entities never 'live alone'.

- Error handling
 I believe this can be simplified as well. Services or data-fetching does not necessarily need to 'try-catch' and check for everything. It's actually a good thing that they don't handle the errors if we use the transaction-handler idea. We could also probably use something like https://github.com/Vincit/db-errors with `onPreResponse`-hapi hook to format and check for different errors, and return a meaningful error message to the client. 

- Uuid and id seperation
[...]

PR #158, opened 2019-11-22
feat: organisation services

Birkbjo:
I've been deep down a rabbit hole of trying out multiple approaches to how to deal with database-results etc. My original branch was starting to get quite far behind, with a lot of random commits that does not make sense anymore, so the result is that I brought some of the ideas to this branch, and if we like it I can introduce the other stuff as well (eg. filtering and paging).

This touches on most of the keypoints in the [issue](#139) raised earlier. I will try to tackle the keypoints and explain my thoughts.

### Services
 We didn't want an ORM and that's fine. However I found using the raw database results very cumbersome. "Service" is a very imprecise term, but it's what I came to call this, but they are not classes or injected in any way - they are simple functions exported together. I'm not opposed to splitting up in different files, but having an index-file or the like to group them so you can do `Organisation.find()` is more developer friendly in my opinion. The services are not fundamentally different from the current "data-consept". They are simple functions that fetch data using knex, but also transform this using the models to an internal model. 

 We can also reuse the data-handlers if we want to extract that code out of the services, but as I tried out filters etc. I found that it was better to be able to access the queries directly instead of executing them elsewhere. 

Keypoints:
- Solves transactions annoyances. Every service-function takes in a db-parameter as the last parameter and can thus be given a transaction to compose other service-functions together in a single transactions. See the tests ([1](https://github.com/dhis2/app-store/pull/158/files#diff-56c19ccaec8e083002b4d7770c1dbce6R152)) for examples of this.
   - An annoying part of this is that it can be easy to forget to pass this parameter, crashing the app. Potential solutions: a default-parameter and set it to the knex-instance, register the service as a plugin (and have access to knex from the context) or just keep it as is. 
- Easier to import and use, less verbose naming. The ideas is that we have common function-names per service, ie `find, create, findByUuid, remove, etc`. Where eg. `find` would accept a filter-object that can be used to construct a filtered-query according to passed query-params.

### Models

I really tried to stay away from classes here, but in many ways it would make sense. In the end I decided to use joi-schemas to describe the object and simple transformations. The idea is that we have three "model-shapes". `Internal`, `database` and `external`. Using [`joi.alter()`](https://hapi.dev/family/joi/?v=16.1.7#anyaltertargets) and [`joi.tailor()`](https://hapi.dev/family/joi/?v=16.1.7#anytailortargets), we can use the same base and describe what changes to schema we want per key. We can also expand upon the schemas using `.append()`.

The idea is that the described object is the "internal" model. This should hopefully be pretty similiar to the external-output. Renames on the internal model (without alterations) are from db -> internal, as this was the use case I wanted solve at first.

The idea of `external` is still work in progress, and some endpoints would probably want other schemas and validations per endpoint, but the idea is that they can use this schema as a base, and either expand or restrict based on that shape, and keep common external rules together (ie strip IDs).

When an object is fetched from the database we call `parseDatabaseJson`. This function can be changed to anything, but for simple transformations like renaming, I've opted for a default function that basically just maps the result through the joi schema.

One downside is that we lock in to joi, however it's very powerful and quite easy to get into (atleast the concepts I've used). 

Keypoints: 

- Joi used for transforming and validating data
- Override functions (`parseDatabaseJson`, `formatDatabaseJson`) if we need more special handling (mapping from id to uuid)
### Error handling

As you can see I don't try to manually catch errors. I believe that catching e.g. database-errors and rethrowing them as a custom error like `Failed to get organisation ${error}` is just unncessary code. I might be lazy, but just letting the native errors bubble up is cleaner in my opinion. I've also worked on a solution to map these errors using [db-errors](https://github.com/Vincit/db-errors) to client-friendly errors.

On another note I believe using [`joi.attempt`](https://hapi.dev/family/joi/?v=16.1.7#attemptvalue-schema-message-options) instead of `joi.validate` is the way to go, as we tend to just check for the error and throw that anyways.

TODO:

- [ ] Add route handler for organisation

This explanation is pretty much longer than the code, so I hope that you look through it and respond with some feedback!

### DIFF
--- a/src/routes/v2/organisations.js
+++ b/src/routes/v2/organisations.js
@@ -1,0 +1,265 @@
+const Boom = require('@hapi/boom')
+const Joi = require('@hapi/joi')
+const {
+    canCreateApp,
+    getCurrentAuthStrategy,
+    getCurrentUserFromRequest,
+} = require('../../security')
+const getUserByEmail = require('../../data/getUserByEmail')
+const { Organisation } = require('../../services')
+const OrgModel = require('../../models/v2/Organisation')
+
+module.exports = [
+    {
+        method: 'GET',
+        path: '/v2/organisations',
+        config: {
+            tags: ['api', 'v2'],
+            response: {
+                schema: Joi.array().items(OrgModel.externalDefintion),
+                modify: true,
+            },
+        },
+        handler: async (request, h) => {
+            const { db } = h.context
+
+            //get all orgs, no filtering
+            //TODO: add filtering
+            const orgs = await Organisation.find({}, h.context.db)
+            return orgs
+        },
+    },
+    {
+        method: 'GET',
+        path: '/v2/organisations/{orgId}',
+        config: {
+            auth: 'token',
+            validate: {
+                params: Joi.object({
+                    orgId: OrgModel.definition.extract('id').required(),
+                }),
+            },
+            tags: ['api', 'v2'],
+            response: {
+                schema: OrgModel.externalDefintion,
+                modify: true,
+            },
+        },
+        handler: async (request, h) => {
+            const { db } = h.context
+            const { orgId } = request.params
+            const organisation = await Organisation.findOne(orgId, true, db)
+            return organisation
+        },
+    },
+    {
+        method: 'POST',
+        path: '/v2/organisations',
+        config: {
+            auth: 'token',
+            validate: {
+                payload: Joi.object({
+                    name: OrgModel.definition.extract('name').required(),
+                }),
+            },
+            tags: ['api', 'v2'],
+            response: {
+                schema: OrgModel.externalDefintion,
+                modify: true,
+            },
+        },
+
+        handler: async (request, h) => {
+            const { db } = h.context
+
+            const { id: userId } = await getCurrentUserFromRequest(request, db)
+
+            const createOrgAndAddUser = async trx => {
+                const organisation = await Organisation.create(
+                    { userId, name: request.payload.name },
+                    trx
+                )
+                const query = await Organisation.addUserById(
+                    organisation.id,
+                    userId,
+                    trx
+                )
+                return organisation
+            }
+
+            const organisation = await db.transaction(createOrgAndAddUser)
+
+            return h
+                .response(organisation)
+                .created(`/v2/organisations/${organisation.id}`)
+        },
+    },
+    {
+        method: 'PATCH',
+        path: '/v2/organisations/{orgId}',
+        config: {
+            auth: 'token',
+            tags: ['api', 'v2'],
+            validate: {
+                payload: Joi.object({
+                    name: OrgModel.definition.extract('name'),
+                    owner: OrgModel.definition.extract('owner'),
+                }),
+                params: Joi.object({
+                    orgId: OrgModel.definition.extract('id').required(),
+                }),
+            },
+        },
+        handler: async (request, h) => {
+            const { db } = h.context
+            const { id: userId } = await getCurrentUserFromRequest(request, db)
+
+            const updateObj = request.payload
+
+            const updateOrganisation = async trx => {
+                const organisation = await Organisation.findOne(
+                    request.params.orgId,
+                    false,
+                    trx
+                )
+                if (organisation.owner !== userId) {
+                    throw Boom.forbidden(
+                        'You do not have permissions to update this organisation'
+                    )
+                }
+                await Organisation.update(organisation.id, updateObj, db)
+                return {
+                    id: organisation.id,
+                }
+            }
+
+            const transaction = await db.transaction(updateOrganisation)
+            return h.response(transaction).code(200)
+        },
+    },
+    {
+        method: 'POST',
+        path: '/v2/organisations/{orgId}/user',
+        config: {
+            auth: 'token',
+            tags: ['api', 'v2'],
+            validate: {
+                payload: Joi.object({
+                    email: Joi.string()
+                        .email()
+                        .required(),
+                }),
+                params: Joi.object({
+                    orgId: OrgModel.definition.extract('id').required(),
+                }),
+            },
+            // response: {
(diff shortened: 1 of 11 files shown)


erikarenhill:
I think we're already so depending on the hapi eco-system that depending on joi is not a deal breaker. We also already use it for input and output validation so it's not a matter of adding yet another dependency but instead use it more.

> Easier to import and use, less verbose naming. The ideas is that we have common function-names per service, ie find, create, findByUuid, remove, etc. Where eg. find would accept a filter-object that can be used to construct a filtered-query according to passed query-params.

I agree that much of this refactoring is sound to make. I also think we should not keep functions such as findByUuuid but as you already mentioned make a more generic find which accepts a filter object (as in your code).

Will need to poke around a bit in the code before giving more feedback

src/routes/v2/organisations.js
```diff
@@ -78,7 +83,7 @@ module.exports = [
     },
     {
         method: 'POST',
-        path: '/v2/organisations/{orgUuid}/add',
+        path: '/v2/organisations/{orgUuid}/addUser',
```
erikarenhill: I believe normally we'd just use verbs for the rest api to describe what happens, and the POST is for add/create, then just use a verb-less url such as  /users on the end instead of /addUser

src/routes/v2/organisations.js
```diff
@@ -37,12 +41,13 @@ module.exports = [
...
-                //   schema: OrgModel.externalDefintion,
+                schema: OrgModel.externalDefintion,
+                modify: true
             },
         },
-        handler: (request, h) => {
+        handler: async (request, h) => {
             const { db } = h.context
-            const organisation = Organisation.findByUuid(orgUuid, true, db)
+            const organisation = await Organisation.findByUuid(orgUuid, true, db)
```
erikarenhill: If we're doing this service layer I'd rather see a generic find such as
`await Organisation.find({ uuid: orgUUid }, true, db)`
or similiar as you've done with the name in the other code, a filter object { filter: { uuid: orgUuid } }

src/services/organisation/index.js
```diff
@@ -0,0 +1,145 @@
...
+            db.ref('u.uuid').as('created_by_user_uuid'),
+            db.ref('u.id').as('created_by_user_id')
+        )
+        .innerJoin(
+            db.ref('users').as('u'),
+            'organisation.created_by_user_id',
+            'u.id'
+        )
+
+async function create({ userId, name }, db) {
```
erikarenhill: why the mix of syntax in function declarations between `const xxx => ... ` and  `async function xxx()` ? I suggest we keep it to the same format throughout the code for all functions no matter if async or not. In most places in the backend here we're using the first variant of those

src/routes/v2/organisations.js
Birkbjo: Right. The reason this was "seperated" from the other `.find()` was the extra code to include users, and as I see we only need that for single organisations.

erikarenhill: A suggestion is that we either fetch them in a micro-service similiar way such as
`const organisation = await Organisation.find(.....)
const users = await User.find({ filter: { organisation: org.uuid } })
return {
     organisation,
     users
}
`

or, we add some "include" possibility as I've seen in some cases other projects use (will add complexity)

example:

`
const organisation = await Organisation.find( { filter: { .... }, include: ["user"] })
`

Personally I'd prefer the first way as it's more straight forward even if the second one would basically do the same internally but would increase our code complexity and increase coupling. What do you think @Birkbjo ?

src/services/organisation/index.js
```diff
@@ -0,0 +1,145 @@
+const joi = require('@hapi/joi')
+const slugify = require('slugify')
+const uuid = require('uuid/v4')
+const { applyFiltersToQuery } = require('../../utils/databaseUtils')
+const { NotFoundError } = require('../../utils/errors')
+const Boom = require('@hapi/boom')
```
erikarenhill: Boom is not used in service layer

src/routes/v2/organisations.js
Birkbjo: Hmm... I would expect at least a reference in the organisation or the user objects. Maybe the Organisation's `.find()` method could call the user.find and append there? Then we would need the include option. Or do we want services to be independent of each other?

It depends on what kind of API we want, but getting a normalized structure like that is not what I would expect from a resource.

src/services/organisation/index.js
Birkbjo: Not really a reason, I think it was just because I extracted it from a function. You would prefer to refactor the other functions to use const instead?

src/routes/v2/organisations.js
Birkbjo: Another reason is that findByUuid always returns one single object (or throws if not exists). And thus nicely maps to the respective endpoints, without the need to check for an empty collection in the case of the `uuid endpoint`. 

You could argue that we shouldn't couple the services to the handlers, and I agree that we should try not to do that, but it's a nice bonus if it's natural mapping. I feel like the distinction between "get a single object by the identifier" and "get a list with this criteria" is distinct enough to have them as different functions.
However I don't really like the "includeUsers" parameter-solution either. Maybe we should move it to an options object or something.

erikarenhill: A suggestion about findByUuid is to rename it to "findOne", "one", "first" or "single" to make it even clearer that it only returns one result, and then we can still use the same filter/config object to pass either uuid, name, or whatnot we want to use in the where-part of the query.

src/services/organisation/index.js
erikarenhill: yeah I think so, it would at least be more consistent with other parts of the project (and I personally think it's more good looking :) )

varl:
I think the comments you had @erikarenhill has been accounted for. Please have another look.

varl:
@Birkbjo this needs to be rebased on master, and before merging I think squashing the commits would be nice.

PR #159, opened 2019-11-25
test: make sure creating an app through post works

erikarenhill:
this PR also contains a minor refactor removing the use of Blob for plain json- data


test/routes/createApp.js
```diff
@@ -0,0 +1,75 @@
...
+const { it, describe, beforeEach, afterEach } = (exports.lab = Lab.script())
+
+const { expect } = require('@hapi/code')
+
+const knexConfig = require('../../knexfile')
+const db = require('knex')(knexConfig)
+
+const { init } = require('../../src/server/init-server')
+
+describe('test create app', () => {
```
varl: This is a great test case to have!

PR #171, opened 2019-12-17
feat: errormapper-plugin

Birkbjo:
Introduces a small plugin that hooks into `onPreResponse`. This is used to map database errors such as "UniqueViolationError" to the correct Boom httpError (ie conflict in this case). We can also add our own errors, such as NotFoundError. It is not restricted to database-errors.
Previously all DB-errors were returned as 500-errors, which is not descriptive enough, as it might indicate user error.

It uses https://github.com/Vincit/db-errors to parse and check for database-errors, and then maps these error-classes to the correct Boom-error.

By default Hapi is stripping all error-messages and returns generic messages as to not leak information. For instance, Joi `ValidationErrors` are returned to the client as `400: Bad Request` with a generic message along the lines of `Invalid request payload input`.
To override this to get more information about the validation (eg what field failed etc.), we would need to re-throw the error in a `failAction`-handler. An idea is to re-throw it in development, and return just the error code with the reason phrase in production. See [this issue](https://github.com/hapijs/hapi/issues/3706) for more information regarding that.

Along those lines there's an option `preserveMessage` that can be set to `true` to include the original database message, which could be toggled in development. We absolutely do not want this enabled in production though, as these are the raw, internal database-errors.

**Note**: this is based on the #158 branch

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6882,6 +6882,11 @@
             "resolved": "https://registry.npmjs.org/dateformat/-/dateformat-3.0.3.tgz",
             "integrity": "sha512-jyCETtSl3VMZMWeRo7iY1FL19ges1t55hMo5yaam4Jrsm5EPL89UQkoQRyiI+Yf4k8r2ZpdngkV8hr1lIdjb3Q=="
         },
+        "db-errors": {
+            "version": "0.2.3",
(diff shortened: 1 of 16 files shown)


Birkbjo:
Closing this in favor of #180. This was started on a branch that was later squashed, so the merge conflicts became way too much to handle.

PR #185, opened 2020-01-13
feat: add possibility to delete a channel in the api

erikarenhill:


src/data/deleteChannel.js
```diff
@@ -0,0 +1,43 @@
...
+            .where('channel_id', uuid)
+            .count()
+        if (+count > 0) {
+            return {
+                deleted: 0,
+                success: false,
+                message:
+                    'Can not delete a channel which has apps published to it.',
+            }
+        }
```
Birkbjo: Instead of doing this you can throw any of the errors from db-errors (see [here ](https://github.com/dhis2/app-hub/blob/master/src/plugins/errorMapper.js#L15), ie. an CheckViolationError, or you can create a custom one). This would remove the need for this object-api, and you could just return the count or something. In the try-catch you would need to rollback and then just rethrow the error as the errormapper will handle the rest.

test/data/channels.js
```diff
@@ -60,4 +61,20 @@ describe('@data::channels', () => {
             'No transaction passed to function'
         )
     })
```
Birkbjo: Would be nice with a test for when the channel has apps.

src/routes/v2/channels.js
```diff
@@ -126,4 +127,41 @@ module.exports = [
...
+            request.logger.info('In handler %s', request.path)
+
+            if (!currentUserIsManager(request)) {
+                throw Boom.unauthorized()
+            }
+
+            const { uuid } = request.params
+            const knex = h.context.db
+
+            try {
```
Birkbjo: There's an potential error here if some error is thrown in the try-block, as the transaction is not rolled back.

PR #186, opened 2020-01-14
fix(plugin): enable error-mapper

Birkbjo:


test/routes/v2/organisations.js
```diff
@@ -285,8 +283,8 @@ describe('v2/organisations', () => {
                 },
             }
             const res = await server.inject(opts)
-            expect(res.statusCode).to.equal(409)
-            expect(res.result.message).to.be.equal('Organisation not found')
+            expect(res.statusCode).to.equal(404)
+            expect(res.result.message).to.be.equal('Not Found')
```
erikarenhill: being picky about the commit message, the fixed test should not belong to "fix(plugin): enable error-mapper" ? But instead go in an own commit.

Birkbjo: I will deduplicate them, but it's related as this would return a 500 error without the error-mapper.

PR #194, opened 2020-01-21
feat: filtering

Birkbjo:
Adds filtering support through a plugin. 

The plugin takes care of the query-parameters and collects this into one `filter`-object in the query. 
The biggest challenge during this was the transformation/renaming of the column-names. I wanted to be able to use Joi-schemas to validate the filters and ideally use the already defined models as a base. I want the consuming functions (service-functions) to not need a lot processing of the filters. However we can't keep all this logic in the plugin, as the service-layer should be usable without it. I wanted the filters follow the idea of internal models - as we do when updating etc  (eg. passing the external/internal name of a column instead of the "database-column name" to the service).

The filter-API is a bit like in DHIS2. You can do `name=eq:DHIS2`. Note that we are not using multiple `filter=` as in DHIS2, as that is a bit weird to me. There are a couple of ways of specifying which fields should be filterable. First we have the `ignoreKeys`-array in the plugin - this can be used to globally ignore keys for all routes, or per routes as well. If Joi-schemas are used to validate the filter, you can use Joi to remove fields that are not specified in the schema (ie. `stripUnknown`). Remember that by default, unknown keys are not allowed - so you would get a ValidationError if a user specified a filter that is not specified/supported. 

To be able to validate the query-object with Joi, we need an regular object - just the key and value (validating of the operator is done when parsing the string). The problem with this approach is that we have to "remove" the operator from the object, and thus we have to "reconcile" it again after successful validation. If the key was renamed during the Joi-validation it's hard to know what the key was renamed to. To solve this we need to insepct the Joi-schema and get the renames from there. This is considered an "expensive" operation, but as our filters shouldn't be big objects this shouldn't really be an issue, I also did a small optimization so it's not called unless a rename is encountered.

To encapsulate this logic I went for a `Filters`-class-design, so that it's usable outside of the plugin as well. I'm still unsure if the singular Filter-objects in the Filters-class should be a class or just regular objects. 

Topics I would like input on:

- I've thought of the idea of supporting [`tag`](https://hapi.dev/family/joi/api/?v=17.1.0#anytagtags) to easily restrict which properties of a Model that is filterable. Currently if we use a Model as a base, all properties in the schema are allowed - but some might not make sense to filter, and may cause issues. A problem is that we don't want to call `schema.describe()` on every route-hit. A possible solution is to scan all routes on startup and save the descriptions, or cache them per route-path. The configuration shouldn't change during runtime anyways.
    - Another simpler idea is to have a `whitelist`, so only those keys are filterable. A problem with this approach is that this means another configuration on filterable-routes, as well as partioning configuration for the filters. It would be nice if it all were contained in a Joi-schema.
    - 

### Filtering referenced models

We need to discuss how to handle related models (starting with only supporting filtering by ID). I've been thinking that we can have this as a regular filter - like we do in DHIS2. This means that the service layer will need have special cases for these "special filters" in the find-function, which is not super clean, and not something that is easy to make generic. However we cannot really make stuff like this generic anyway, since we don't use an ORM. It can also be argued that the filter should have a smaller scope by handling "searching only". 

After a chat with @erikarenhill he proposed that we could use "resource-URLs" instead. At the moment we only have one implemented example of this kind of filtering; "get all organisations that some user(id) is a member of". Currently that would look like `/v2/organisations?user:eq:SOMEID`, with a resource-URL it would look like `v2/users/SOMEID/organisations/`. To me this seems more complicated. We need to introduce a new `users` endpoint, and we need to decide where the logic for fetching this should be - users-layer or organisation-layer? Should it return a list of ID, or an array of the objects? We would also need to introduce new endpoints for every new entity, basically moving the responsibility of filtering to away from the code that deals with that particular entity. I guess it's more "restful" though.

At the moment I'm still leaning towards the filtering solution. It's quite similar to DHIS2, it's seems simpler and will probably require less code.


Birkbjo:
Since last time I've digged deep into Joi-code to understand how [extensions](https://hapi.dev/family/joi/api/?v=17.1.0#extensions) work, as they are not documented. 

With that I've implemented a new Joi-type for filters. The idea  is that instead of having the queryFilter validate queries, we use the built in 'query'-validation in Hapi (with Joi). The `filter`-type can parse strings of `operator:value`, which results in an object with these properties (we might want to use the Filter-class).
A cool benefit of this is that we can use all the powerful features of Joi to validate the values of the Filter. We can do stuff like `Joi.filter(Joi.string().guid())`, to only allow guids as the value of the filter. Note that `Joi.filter(joiSchema)` is a shortcut for `Joi.filter().value())`. We can also do `Joi.filter().operator(Joi.valid('eq'))` to specify allowed operators. The defaults are Joi.string() for value and 

I've decided to not do the transformation of API-operators to DB-operators and renames inside the extension. I want the scope of the extension to be rather small and to follow Joi-patterns. So no custom transformation inside the extension-code. This is kind of conflicting to my initial thought of re-using the Models to use as query validations, however I realized that it would cause more problems than it was worth due to swagger etc.
You could in theory use stuff like `.rename()`, but this would generate confusing Swagger-output (as you need to describe both the DB-column and the field-name in the Joi-schema - and that would show up in Swaggers output). 

I'm still not completely done yet, as I need to figure out a nice way to map field-names to column names and operator to DB-operators. My idea is that this will be done in the `Filters`-class, as you should be able to `getFilter()` by the internal/external name of the filter - so it needs to be aware of both the DB-column name and internal/external name. It will probably support both a regular object-map and a Joi-schema, so we can pass in a Model and use those renames directly.

varl:
Good update, thanks for that @Birkbjo. Really nice that you mention the new discoveries lead to a shift away from the original plan. :+1:

erikarenhill:
@hapi/bounce used in queryFilter.js is missing as dependency in package.json

server/src/plugins/queryFilter.js
```diff
@@ -0,0 +1,142 @@
...
+    const routeQueryValidation = request.route.settings.validate.query
+    const rename = routeOptions.rename
+    const queryFilterKeys = new Set()
+    let renameMap = rename
+    let validateDescription = validateDescriptions[request.path]
+    let renameDescription = renameDescriptions[request.path]
+
+    if (rename) {
+        if (Joi.isSchema(rename)) {
+            if (!renameDescription) {
```
erikarenhill, line 67: personally I'd add these to the same if-case with && in between to avoid nesting scopes so deep as there are no else cases

server/src/plugins/queryFilter.js
```diff
@@ -0,0 +1,142 @@
...
+    }, {})
+
+    try {
+        const filters = new Filters(queryFilters, {
+            renameMap,
+        })
+        request.plugins.queryFilter = filters
+    } catch (e) {
+        Bounce.rethrow(e, 'system')
+        throw Boom.boomify(e, { statusCode: 400 })
```
erikarenhill, line 121: Wouldn't the Bounce.rethrow cause the Boom.boomify row to never run?

server/src/routes/v2/organisations.js
```diff
@@ -1,5 +1,5 @@
 const Boom = require('@hapi/boom')
-const Joi = require('@hapi/joi')
+const Joi = require('../../utils/CustomJoi')
 const {
     canCreateApp,
```
erikarenhill, line 4: not used in organisations?

server/src/plugins/queryFilter.js
Birkbjo: No, Bounce just rethrows system errors, eg. syntaxErrors. It's so that we don't throw 400-errors for errors that really should be 500-errors.

erikarenhill: right, so when there is a system error this would throw in all other cases it would just skip and the Boom exception will be executed, I understand

Birkbjo:
> We need to get swagger to somehow generate docs about available filter parameters

I tested the swagger output, and it does list the filter parameters listed in `validate.query`. However it will only show the type as strings, which seems to be the case path-params as well, so that is not related to the custom Joi extension. 
So we could have some documentation about how the filtering works and the available operators somewhere else, and show the available filter-keys per route. We could also use `Joi.description()` to describe the parameters per route.

server/src/routes/v2/organisations.js
```diff
@@ -17,18 +15,29 @@ module.exports = [
         config: {
             tags: ['api', 'v2'],
             response: {
-                //  schema: Joi.array().items(OrgModel.externalDefintion),
-                //modify: true,
+                schema: Joi.array()
+                    .items(
+                        OrgModel.externalDefintion.fork('users', s =>
```
erikarenhill, line 20: incorrect spelling of definition

PR #198, opened 2020-01-24
fix: users-with-organisation view

Birkbjo:
Small fix for the view. The view is not used by anything yet, but is handy to have.
I believe it's fine to update migrations directly like this when it's a bug-fix right?

### DIFF
--- a/migrations/20200203182043_fix_view_users_with_organisations.js
+++ b/migrations/20200203182043_fix_view_users_with_organisations.js
@@ -1,0 +1,23 @@
+exports.up = async knex => {
+    await knex.raw('DROP VIEW users_with_organisations')
+    await knex.raw(`
+        CREATE VIEW users_with_organisations  
+        AS SELECT u.id AS user_id, u.name as user_name, u.email, o.name AS organisation_name, o.id AS organisation_id FROM users AS u 
+            INNER JOIN user_organisation AS dorg 
+                ON dorg.user_id = u.id
+            INNER JOIN organisation AS o 
+                ON o.id = dorg.organisation_id
+        `)
+}
+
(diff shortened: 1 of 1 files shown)


migrations/20200203182043_fix_view_users_with_organisations.js
```diff
@@ -0,0 +1,13 @@
...
+        CREATE VIEW users_with_organisations  
+        AS SELECT u.id AS user_id, u.name as user_name, u.email, o.name AS organisation_name, o.id AS organisation_id FROM users AS u 
+            INNER JOIN user_organisation AS dorg 
+                ON dorg.user_id = u.id
+            INNER JOIN organisation AS o 
+                ON o.id = dorg.organisation_id
+        `)
+}
+
+exports.down = async function(knex) {}
```
erikarenhill: need to implement the down step as well

Birkbjo: Wouldn't this break stuff? It's already dropped in the rollback of the first migration that created the view. So while doing a rollback it would try to drop the table multiple times.

erikarenhill: This is a view and not a table, and it will break not having a rollback since then the database will contain a view that the code might not support

erikarenhill: The rollback should drop the view, and recreate the view as it was before being dropped in the up-state, e.g. just copy paste the previous closest migration creating the view, and then drop + paste that in the down step and you'll be fine

Birkbjo:
> isn't the up and down view the same?

No, notice line 7 and 19. The reference to u.id is wrong.

PR #215, opened 2020-02-03
fix(errormapper): expose 500-errors when preserveMessage is true

Birkbjo:


src/plugins/errorMapper.js
```diff
@@ -10,7 +10,7 @@ const {
     ConstraintViolationError,
     wrapError,
 } = require('db-errors')
-
+const debug = require('debug')('apphub:plugins:errorMapper')
```
erikarenhill: I would opt for using scope `apphub:server:plugins:errorMapper` (add server after apphub) as we try to keep server and client in separate debug scopes

Birkbjo: They wouldn't be in the same log though so to me it's a bit overkill, but I will comply!

Birkbjo: Nvm, I don't actually need to use debug here at all.

PR #224, opened 2020-02-11
chore(ci): build and push to dockerhub

erikarenhill:
This does a full build with docker on all pull requests, will also push a docker image to dockerhub registry on changes to master.


.github/workflows/dockerpush.yml
```diff
@@ -0,0 +1,99 @@
+name: Docker
```
varl: `name: 'DHIS2: Docker'`

.github/workflows/dockerpush.yml
```diff
@@ -0,0 +1,99 @@
...
+    needs: test
+
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build image
+        run: docker build . --file Dockerfile --tag image
```
varl: Wouldn't this name the image `image`?

erikarenhill: yes, but a couple of rows below, line 98 (https://github.com/dhis2/app-hub/pull/224/files#diff-7d3075cfc8d4fd3a38fa85f7abe0e381R98)  it re-tags it `docker tag image $IMAGE_ID:$VERSION` as taken from githubs example

PR #225, opened 2020-02-11
feat: show Joi-errors in development

Birkbjo:


Birkbjo:
The failActions are mostly defined on the `response`-part, which is not the same. By throwing the error in the failAction we will get more specific validation information, instead of a generic error message. An example of such a response: 
```
{
    "statusCode": 400,
    "error": "Bad Request",
    "message": "\"asf\" is not allowed",
    "validation": {
        "source": "query",
        "keys": [
            "asf"
        ]
    }
**}**
```
This would indeed be overridden by route-validation-failHandlers.

Regarding the defaultFailAction, this will always return generic 500-errors, which is fine if we only use them in 'response'-validations - as that is actually a developer error. What is nice about this `global` failHandler is that you will get the detailed error in the response as well, instead of having to look in the server logs.

PR #227, opened 2020-02-12
fix: correct Dockerfile so it builds

erikarenhill:
I accidentally merged an incorrect Dockerfile when squashed other commits and lost some changes.. this is the correct one that builds.


.github/workflows/dockerpush.yml
```diff
@@ -43,7 +52,7 @@ jobs:
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     steps:
       - uses: actions/checkout@v2
-      - name: install deps
+      - name: install deps for tests
```
varl: Capitalize the name: Install deps for tests.

PR #238, opened 2020-02-20
refactor: change to workspaces structure

varl:
This enables separation between server and client dependencies, which
makes it easy to know what is used where.

It also reduces the footprint of the final Docker image, as we only need
the dev deps when building the client bundle, so in the final image we
can include only the server runtime deps.

# TODO

- [x] Go over the test commands
- [x] Setup concurrently for running client and server
- [x] Figure out what to do with: https://github.com/dhis2/app-hub/blob/master/.github/workflows/dockerpush.yml#L37

Anything else you know we need?


varl:
https://github.com/dhis2/docker-compose/pull/15 needs to be merged after this PR.

PR #248, opened 2020-02-26
fix: add required validation to NewAppVersion form

erikarenhill:
add required form validation to "create new version" - to not crash when trying to submit with incorrect values in the POST to the backend

### DIFF
--- a/client/src/components/form/NewAppVersionForm.jsx
+++ b/client/src/components/form/NewAppVersionForm.jsx
@@ -15,7 +15,7 @@ const DHISVersions = config.ui.dhisVersions
 
 const validate = values => {
     const errors = {}
-    const requiredFields = ['version', 'file', 'channel']
+    const requiredFields = ['version', 'file', 'channel', 'minVer', 'maxVer']
     requiredFields.forEach(field => {
         if (!values[field]) {
             errors[field] = 'Required'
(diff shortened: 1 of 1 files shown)


client/src/components/form/NewAppVersionForm.jsx
```diff
@@ -15,7 +15,7 @@ const DHISVersions = config.ui.dhisVersions
 
 const validate = values => {
     const errors = {}
-    const requiredFields = ['version', 'file']
+    const requiredFields = ['version', 'file', 'channel', 'minVer', 'maxVer']
```
varl: Questions:

- Is channel hardcoded to "stable" in the client?
- Are minVer and maxVer not possible to send empty strings? so e.g. minVer of 2.33 without a maxVer is not valid?

erikarenhill: Available channels are loaded dynamically from the backend.

This validation just adds up to be in line with the same requirements as when creating a new app.

Regarding max version maybe that requires a discussion how we want to handle it before implementing. What I can tell from the API we allow empty max but the database doesn't. 

I wanted to bring the applications validation to a consistent working state before touching that area.

varl: OK, if there's a plan to account for it I'm OK with this.

PR #269, opened 2020-02-28
fix: reduce docker context building locally

erikarenhill:
modify dockerignore due to workspace changes to include node_modules from subdirectories for example


varl:
Works nicely, seeing a reduction from 83 MB to 11 MB.

PR #273, opened 2020-03-03
feat: add support to clone all approved apps between two app hub instances

erikarenhill:
Found some bugs while scripting a clone of the old app store to new app hub

Added a clone script which can download approved apps from an existing app hub and post/create to another one.

Todo:
Unit tests for the found and fixed bugs


tools/.env.template
```diff
@@ -0,0 +1,11 @@
+#Required to use the clone script
```
varl: As a general note, it is helpful if template files are not hidden so the user can see that there is a template available.

erikarenhill: renamed it

PR #285, opened 2020-03-11
refactor: split media to own table and connect to app instead of version

erikarenhill:
Since this PR also changes the way files are stored on disk with naming (previously stored in appversion id subdirectory) this is to be considered a breaking change.

Should with other words be pushed to staging and prod before going live and adding real apps there.


server/src/data/addMedia.js
```diff
@@ -0,0 +1,104 @@
...
+const { ImageTypes } = require('../enums')
+
+const paramSchema = joi
+    .object()
+    .keys({
+        userId: joi
+            .string()
+            .uuid()
+            .required(),
+        imageType: joi
```
varl, line 12: We call it `appMedia`, so maybe we could make this more generic as well. For example, it would be possible to upload a video show case of the app as app media.

erikarenhill: True, but calling it appMediaType could be confusing as well because we also have a mediaType (mime) table.

varl: Just `type`? It's in the context of the `AppMedia` so being generic here should be fine.

varl: Or `mediaType`?

erikarenhill: Maybe its more appropriate to name this mediaType and then rename the existing mediaType to mimeType instead?

varl: Agreed, better to rename to `mediaType` and `mimeType`. Good point about `type` being reserved as a keyword as well.

erikarenhill: I don't think it is reserved in javascript, but I usually avoid it since it often is in other languages. Anyhow, I'll rename it as we agree on this!

PR #307, opened 2020-04-08
fix: allow admins to download all apps

erikarenhill:
this adds possibility to authenticate the user on the download endpoint for apps, and passes the JWT-token in the querystring on download links when viewing an unapproved/rejected app in the admin/edit-version of the gui.

This also brings in some changes to how a developer is determined to have access to an app. Previously it would only allow edit-access if being the developer that uploaded the app, now it will allow access to all apps which you belong to the same organisation it's connected to.

Also tested that you can retreive "myapps" with a user having multiple organisations

Adds transaction/rollback beforeEach/afterEach organisation unit tests

### DIFF
--- a/client/src/components/appVersion/VersionListEdit.jsx
+++ b/client/src/components/appVersion/VersionListEdit.jsx
@@ -1,3 +1,4 @@
+// eslint-disable-next-line react/no-deprecated
 import React, { Component, PropTypes } from 'react'
 import { connect } from 'react-redux'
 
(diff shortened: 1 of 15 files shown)


server/src/routes/v1/apps/handlers/getAppFile.js
```diff
@@ -1,14 +1,18 @@
+const Boom = require('@hapi/boom')
+
 const debug = require('debug')('apphub:server:routes:handlers:v1:getAppFile')
 const { AppStatus } = require('../../../../enums')
 const { getFile } = require('../../../../utils')
 
+const { currentUserIsManager } = require('../../../../security')
+
 module.exports = {
     //unauthenticated endpoint returning the approved app for the specified id
```
Birkbjo, line 10: Comment is outdated

Birkbjo, line 10: Comment is outdated

client/src/components/appVersion/VersionListEdit.jsx
```diff
@@ -210,11 +210,18 @@ class VersionListEdit extends Component {
 
         const values = merge(version, this.state.editedValues[version.id])
 
+        //auth0 stores the JWT token in localStorage
+        //as only authenticated users can edit an app, just assume this exists in this component
+        const token = localStorage.getItem('id_token')
```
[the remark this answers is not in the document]
erikarenhill: of course, forgot we had this!

server/src/routes/v1/apps/handlers/getAppFile.js
erikarenhill: removed

PR #310, opened 2020-04-22
fix: avoid creating duplicates when cloning

erikarenhill:


tools/clone.js
```diff
@@ -22,21 +32,56 @@ async function main() {
...
+
         if (!fs.existsSync('./apps')) {
             fs.mkdirSync('./apps')
         }
 
-        const dir = './apps/' + app.name
+        const dir = './apps/' + app.name.replace('/', '-')
         if (!fs.existsSync(dir)) {
-            fs.mkdirSync(dir)
+            fs.mkdirSync(dir, '0777', true)
```
varl, line 76: Isn't world writable a bit much? I would expect `755` to be plenty.

Birkbjo: `mkdirSync` also takes a function and not 3 params? So this shouldn't do anything other than fallback to default.

erikarenhill: @varl the default mode is actually 777 so this would only make it more explicit, but of course we can go with 755 instead.

@Birkbjo you are correct, I remembered incorrectly that a third parameter would be recursive creation, but it's passed in a config object. Will change.

PR #316, opened 2020-04-22
fix(clone): keep owner access

Birkbjo:
Note that this builds upon #310 , so that should probably be merged first.

To keep access for the uploading user we need to add that user to the organisation the app is uploaded to. The developer-user and owner may be completely different emails/users. Normally, when uploading an app, the user uploading the app are the owner, and thus added to that organisation (fixed in #314 ). However when uploading apps with the clone script, the M2M user will be the current user, and therefore the user that uploaded the app in the previous appstore won't have access unless the developer email is the same.

The fix is not pretty to be honest. I've modified the v1 API, but this is not breaking as it just adds a new object that can be attached to the app. This object describes the owner of the app, so it can "override" the current user as the owner of the app with the attached information.

The clone script uses the M2M "API explorer client" to get the information about the original uploader from Auth0 (using the owner-field that has the oauth ID for the original uploader). 

A pretty useful utility when testing access for different users is to go to `Users` in Auth0, and select `Sign in as user`. Click advanced and add `openid roles user_id email profile` as Scope. You can even use localhost as the callback-url.

If you are testing this, you need to add the `AUTH0_MANAGEMENT_SECRET`, `AUTH0_MANGAGEMENT_AUDIENCE` and `AUTH0_MANAGEMENT_CLIENT_ID` keys in the .env-file in the tools directory. The values are found in auth0 under the `API Explorer Client` application.

### DIFF
--- a/server/src/models/v1/in/CreateAppModel.js
+++ b/server/src/models/v1/in/CreateAppModel.js
@@ -6,9 +6,7 @@ const CreateModelAppData = Joi.object().keys({
     name: Joi.string(),
     description: Joi.string().allow(''),
     appType: Joi.string().valid(...AppTypes),
-    sourceUrl: Joi.string()
-        .uri()
-        .allow(''),
+    sourceUrl: Joi.string().uri().allow(''),
     developer: Joi.object().keys({
         name: Joi.string(),
         email: Joi.string().email(),
@@ -30,6 +28,12 @@ const CreateModelAppData = Joi.object().keys({
             description: Joi.string().allow('', null),
         })
     ),
+    owner: Joi.object({
+        email: Joi.string()
+            .email()
+            .required(),
+        name: Joi.string().required(),
+    }),
 })
 
 const payloadSchema = Joi.object({
--- a/server/src/routes/v1/apps/handlers/createApp.js
+++ b/server/src/routes/v1/apps/handlers/createApp.js
@@ -177,7 +177,7 @@ module.exports = {
                 //1. either it's a manager uploading the app
                 //2. or the owner e-mail is the same as verified on the request
                 const shouldAddUserToOrg =
-                    isManager || appOwner.email === currentUser.email
+                    isManager || email === currentUser.email
 
                 if (shouldAddUserToOrg && appOwner === null) {
                     appOwner = await createUser(
--- a/server/src/routes/v1/apps/handlers/createApp.js
+++ b/server/src/routes/v1/apps/handlers/createApp.js
@@ -19,7 +19,7 @@ const createAppVersion = require('../../../../data/createAppVersion')
 const createLocalizedAppVersion = require('../../../../data/createLocalizedAppVersion')
 const addAppVersionToChannel = require('../../../../data/addAppVersionToChannel')
 const addAppMedia = require('../../../../data/addAppMedia')
-
+const OrganisationService = require('../../../../services/organisation')
 const {
     getOrganisationsByName,
     createOrganisation,
@@ -162,15 +162,15 @@ module.exports = {
             }
 
             const organisationId = organisation.id
-            let ownerUserId = currentUserId
+            const requestUserId = currentUserId
             const developerUserId = appDeveloper.id
 
             if (appJsonPayload.owner) {
                 const { email, name } = appJsonPayload.owner
-                const appOwner = await getUserByEmail(email, knex)
+                let appOwner = await getUserByEmail(email, knex)
 
-                if (appOwner == null) {
-                    const ownerUser = await createUser(
+                if (appOwner === null) {
+                    appOwner = await createUser(
                         {
                             email,
                             name,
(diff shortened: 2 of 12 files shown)


server/src/routes/v1/apps/handlers/createApp.js
```diff
@@ -154,13 +154,40 @@ module.exports = {
...
             const organisationId = organisation.id
-            const requestUserId = currentUserId
+            let ownerUserId = currentUserId
             const developerUserId = appDeveloper.id
 
+            if (appJsonPayload.owner) {
+                const { email, name } = appJsonPayload.owner
+                const appOwner = await getUserByEmail(email, knex)
+
+                if (appOwner == null) {
```
erikarenhill, line 164: should use strict equality?

server/src/routes/v1/apps/handlers/createApp.js
```diff
@@ -154,13 +154,40 @@ module.exports = {
...
+                } else {
+                    ownerUserId = appOwner.id
+                }
+            }
+
             //Create the basic app
             const dbApp = await createApp(
                 {
-                    userId: requestUserId,
+                    userId: ownerUserId,
```
erikarenhill, line 190: created by id should still be the request user id, If i'm as admin add an app, with another owner and developer, we should still be able to track that it was me that did this action

erikarenhill: and the same applies to the below changes

server/src/routes/v1/apps/handlers/createApp.js
Birkbjo: I normally use strict yeah, but this catches undefined as well. I can change it though

server/src/routes/v1/apps/handlers/createApp.js
```diff
@@ -165,6 +165,46 @@ module.exports = {
...
+                if (appOwner === null) {
+                    appOwner = await createUser(
+                        {
+                            email,
+                            name,
+                        },
+                        knex,
+                        trx
+                    )
+                    await addUserToOrganisation(
```
erikarenhill, line 181: this pose a security risk as you could add whatever user you want to any organisation and hijack that organisation as long as the user did not already exist (e.g. create a new user by passing a new e-mail in the owner object pointing to an existing organisation)

server/src/services/organisation.js
```diff
@@ -133,6 +133,17 @@ const remove = async (id, db) => {
         .delete()
 }
 
+const hasUser = async (id, userId, knex) => {
+    const hasUser = await knex('user_organisation')
+        .select(1)
```
erikarenhill, line 138: I would use the native SQL count here instead of selecting 1 and check for result length.

http://knexjs.org/#Builder-count

server/src/models/v1/in/CreateAppModel.js
```diff
@@ -30,6 +28,12 @@ const CreateModelAppData = Joi.object().keys({
             description: Joi.string().allow('', null),
         })
     ),
+    owner: Joi.object({
+        email: Joi.string()
+            .email()
+            .required(),
+        name: Joi.string().required(),
+    }),
```
erikarenhill, line 36: owner should be optional? Otherwise we must also send this from the web client when uploading new apps.

Birkbjo: This is not required? It's optional by default, the required is just when the object is given, it is required to have email and name.

server/src/services/organisation.js
Birkbjo: Why? To my knowledge this is more performant, and the way I've always done "checks" like these. Not that it really matters here though, but why use an aggregate function when you don't have to? I would rather select id or something and check the result length.

server/src/routes/v1/apps/handlers/createApp.js
Birkbjo: I agree, good catch! I will look into this.

server/src/services/organisation.js
erikarenhill: Maybe you are thinking of count(1)? That's more performant.

server/src/models/v1/in/CreateAppModel.js
erikarenhill: not needed, I remembered incorrectly that objects on the top-level were required unless explicit optional

server/src/services/organisation.js
erikarenhill: It also can come down to where it's used, consider you have a table with 50 million of rows, you wouldn't want to return 50 million ids and then count them in code

erikarenhill: I now see you use the limit(1) on the end as well so that would never been the case.

server/src/routes/v1/apps/handlers/createApp.js
erikarenhill: I think https://github.com/dhis2/app-hub/pull/316/commits/ea6ce0c78e7b03bb337190e8db751287708e21ad will fix this. Please check

server/src/routes/v1/apps/handlers/createApp.js
```diff
@@ -134,9 +135,20 @@ module.exports = {
                 )
             } else {
                 organisation = organisations[0]
-                if (currentUserId !== organisation.created_by_user_id) {
-                    //should we allow anyone to create an app for an existing organisation?
-                    //throw Boom.unauthorized()
+
+                const org = await OrganisationService.findOne(
```
Birkbjo, line 139: Any reason to not use the `hasMember()`-function?

erikarenhill: Not really, copied from organisation handlers where it's done that way, but there we also use other properties which makes more sense to fetch the entire org. Replaced with hasUser

Birkbjo: If you need the org I guess it makes sense with the original way, so you don't need 2 database requests.

PR #337, opened 2020-05-15
feat: organisation management

Birkbjo:


erikarenhill (changes requested):
Multiple issues found:

Clicking cancel on the "Add Member" popup with an empty e-mail makes the required validation show up instead of closing the popup.

When adding a new member to an organisation, the popup does not close

The "pencil"/edit icon up to the right next to the organisation name does not do anything

UX inconsistency; adding an organisation the + sign is below the list to add items to, but the member list has the + sign above the list

As a non manager - trying to create an organisation that already exists, for example DHIS2 makes the SQL error from the backend show up in the snackbar message.

As a non manager - trying creating an application that already exists, even though it gives the sql error above, it creates an organisation with that name and adds the user to it, if the naming differs by casing or white-spaces.

No success message or reload of organisations after creating an organisation (e.g. have to reload the page manually), reload only works for managers/admins, not users

As a non-manager when creating a new application, I would expect to only see my own organisations in the select-list with organisations instead of all available.

erikarenhill:
After testing some more after the most recent commits I find that the prevention of creating similiar (names ending up with the same slug) works well.

For non-managers, the list of organisations still doesn't reload after adding a new one (have to manually reload the page for it to show up)

Birkbjo:
>  Maybe it would be better if you apply to create an organisation and then an admin must approve it instead of being able to create N organisations freely?

Think this will be a bit complicated for now. If we see that it's becoming a a problem we can start looking at that, but at the moment don't really think we need it. Remember that these organisations won't really be visible unless it's linked to an approved app. So other than someone trying to maliciously "occupy" organisation names I don't see this being an issue.

>For non-managers, the list of organisations still doesn't reload after adding a new one (have to manually reload the page for it to show up)
This should be fixed now

erikarenhill:
@Birkbjo 
Refreshing verified working now as non admin!

Still there is a pencil/edit icon on the organisation detail page, either we should remove it until it actually does something, or implement the edit functionality.

erikarenhill (changes requested):
Remove pencil/edit icon if it does not do anything (organisation detail page)

No validation for maximum characters in place (100), only throws an sql error/ 500 internal error. I suppose this is something thats rarein real world would to hit the limit for, just something I noticed/tested.

Trying to upload an app to an org you don't belong to show this message "You are not a member of this organisation. Contact the owner to get access to upload apps.", maybe we should put an e-mail or name there so they know whom to contact?

Birkbjo:
> Remove pencil/edit icon if it does not do anything (organisation detail page)

I somehow didn't manage to push this change...

> No validation for maximum characters in place (100), only throws an sql error/ 500 internal error. I suppose this is something thats rarein real world would to hit the limit for, just something I noticed/tested.

I can add this easily, is that for org-name?

> Trying to upload an app to an org you don't belong to show this message "You are not a member of this organisation. Contact the owner to get access to upload apps.", maybe we should put an e-mail or name there so they know whom to contact?

Yeah Ive thought about that also, I can try to show the developer email. On that note, I think that should be renamed to "contact email" or something in that vein.

erikarenhill:
> > Remove pencil/edit icon if it does not do anything (organisation detail page)
> 
> I somehow didn't manage to push this change...
The popup now shows up, but nothing happens when clicking either Cancel or OK
The console show that the form was submitted when clicking OK, but no request for the actual update are made and the dialog doesn't close.

> 
> > No validation for maximum characters in place (100), only throws an sql error/ 500 internal error. I suppose this is something thats rarein real world would to hit the limit for, just something I noticed/tested.
> 
> I can add this easily, is that for org-name?
Yep!

> 
> > Trying to upload an app to an org you don't belong to show this message "You are not a member of this organisation. Contact the owner to get access to upload apps.", maybe we should put an e-mail or name there so they know whom to contact?
> 
> Yeah Ive thought about that also, I can try to show the developer email. On that note, I think that should be renamed to "contact email" or something in that vein.

erikarenhill (approved):

varl:
Looks good overall, it's quite large and hard to review on GitHub so we probably want to run this a bit in `next` before releasing it to `master`.

Nice work, @Birkbjo !

varl (approved):

PR #339, opened 2020-05-29
chore: use org secrets in docker-publish workflow

amcgee:
@erikarenhill @Birkbjo I added Docker Hub secrets to our org and authorized this repo, we should also maybe move the AWS tokens there?  Leaving them as-is for now


erikarenhill:
@amcgee changing base to next (staging) to prevent conflict when merging next->master with currently unpublished changes

varl (approved):

PR #343, opened 2020-06-02
fix: demourl validation

erikarenhill:
https://jira.dhis2.org/browse/HUB-81

### DIFF
--- a/client/src/components/appVersion/VersionListEdit.jsx
+++ b/client/src/components/appVersion/VersionListEdit.jsx
@@ -25,6 +25,24 @@ import { loadChannels } from '../../actions/actionCreators'
 
 import { getAuth } from '../../utils/AuthService'
 
+import { validateURL } from '../form/ReduxFormUtils'
+
+const validate = values => {
+    const errors = {}
+    const uriFields = ['demoUrl']
+
+    uriFields.forEach(field => {
+        if (values[field]) {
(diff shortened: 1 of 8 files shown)


erikarenhill:
> I'm not as familiar with this codebase, but from an outside perspective the code looks good to me - would be great if @Birkbjo could also take a look.
> 
> @erikarenhill did you consider updating the read API as well to fail gracefully rather than blowing up when an invalid value exists?

@amcgee  I did give that a thought, but at the same time we have the output-validation so it should fail when an invalid value exists, otherwise we could as well remove the entire validation. I could not really make up my mind here and left that as is. A valid point for discussion though

Birkbjo:
> > I'm not as familiar with this codebase, but from an outside perspective the code looks good to me - would be great if @Birkbjo could also take a look.
> > @erikarenhill did you consider updating the read API as well to fail gracefully rather than blowing up when an invalid value exists?
> 
> @amcgee I did give that a thought, but at the same time we have the output-validation so it should fail when an invalid value exists, otherwise we could as well remove the entire validation. I could not really make up my mind here and left that as is. A valid point for discussion though

Yeah, I actually think we could disable output validation in production and potentially enable it in staging. Ie. setting https://hapi.dev/tutorials/validation/?lang=en_US to `log`, so we are logging the errors instead of throwing a 500-error.

PR #361, opened 2020-08-25
fix: use slug + version as filename on download

erikarenhill:
should resolve https://jira.dhis2.org/browse/DHIS2-9256 if the backend makes use of the content-disposition filename from header


erikarenhill:
looks like the core doesn't respect that header

Birkbjo (approved):

erikarenhill:
looks like the core ignore the header, need to change the filename in the url instead.

Birkbjo (approved):

amcgee:
@erikarenhill nice work!  Should we remove the header since it's not used?

PR #363, opened 2020-08-25
fix: verify developer permission correctly on several API endpoints

erikarenhill:
previously a developer (non manager) could not delete an app-version, upload image, or delete an image on his own apps after it was created the first time. This PR fixes all those

https://jira.dhis2.org/browse/HUB-84
https://jira.dhis2.org/browse/HUB-85
https://jira.dhis2.org/browse/HUB-86


Birkbjo (approved):
LGTM

PR #367, opened 2020-08-26
fix(version): use select-field when editing version

Birkbjo:
Adds consistent Version input. Before this was free-text, but its now a Dropdown.


erikarenhill:
Would it be possible to also add drop-downs/selects for the "New version" popup? That's the last place I think which allows text

erikarenhill (approved):
I tested the edit-version case and that's working, but see previous comment for also adding it to "new version"

erikarenhill:
we should also verify that minimum version is less/equal to maximum version. Currently it's possible to add a minimum version lower than the max

PR #370, opened 2020-09-03
fix(apiroutes): add 404-handler for /api

Birkbjo:
Adds a catch-all route for `/api` that returns the default `404-error`. Previously the  catch-all route in staticFrontendRoutes would also catch 404 under `/api`, and return `index.html`. 

http://localhost:3000/api/appz -> would previously return `index.html`. Now it returns 404.


amcgee (approved):
LGTM 🎉

PR #373, opened 2020-09-24
fix(download): encode download-components, forbid _ in version

Birkbjo:
Using `_` in the app-version causes problems as `_` is used to separate the app-slug and the version. Underscores are not encoded by `encodeURIComponent`, so using that will not prevent the invalid URL. 

I've added a migration that replaces all versions with `_` to be replaced with `.`. I've also added validation to the server to prevent versions containing `_`. However we probably should have some more validation to enforce versioning best-practices. Any input on that? We could use `semver` to validate versions, and we do not necessarily need to apply that retroactively.

### DIFF
--- a/server/migrations/20200924140557_version-replace-underscore.js
+++ b/server/migrations/20200924140557_version-replace-underscore.js
@@ -1,18 +1,12 @@
-
 exports.up = function(knex) {
-
     return knex.raw(
         `update app_version
         set version = replace(version, '_', '.')
         where version like '%$_%' ESCAPE '$'
-    `)
-};
+    `
+    )
+}
 
 exports.down = function(knex) {
-
-    return knex.raw(
-        `update app_version
-        set version = replace(version, '.', '_')
-        where version like '%$.%' ESCAPE '$'
-    `)
-};
+    return Promise.resolve()
+}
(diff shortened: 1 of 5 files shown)


server/migrations/20200924140557_version-replace-underscore.js
```diff
@@ -0,0 +1,18 @@
...
+        where version like '%$_%' ESCAPE '$'
+    `)
+};
+
+exports.down = function(knex) {
+
+    return knex.raw(
+        `update app_version
+        set version = replace(version, '.', '_')
+        where version like '%$.%' ESCAPE '$'
```
amcgee, line 16: Technically this isn't correct for the `down` migration - it will convert all versions which were originally `.` to `_` as well, which would be bad.  Can we just omit the `down` migration for this one?

amcgee (changes requested):

Birkbjo:
You are right, fixed. What about semver-validation? I have a patch for it if we want that in as well, or we could do it in a separate PR.

amcgee (approved):
LGTM - I think we should do SemVer validation, yeah.  We could do that in a separate change if you want

Birkbjo:
Will do SemVer in another PR, as it does contain some code change and update to tests. I think it makes sense to separate these.

PR #375, opened 2020-09-28
feat(validation): add semver validation for version

Birkbjo:
Adds Semantic Version validation on app-versions. This is not retroactive, and is only checked for new apps and versions. Validation is done both client and server-side. 

Very open for feedback with regards to the client error-message.


client/src/components/form/ReduxFormUtils.jsx
```diff
@@ -204,3 +205,26 @@ export const validateURL = value => {
...
+        <div style={{ whiteSpace: 'pre-wrap' }}>
+            Must be a semantic version. Use the form <span>x.x.x.</span>
+            <br />
+            <a
+                href="https://docs.npmjs.com/about-semantic-versioning"
+                target="_blank"
+            >
+                Click here
+            </a>{' '}
+            for more information
```
amcgee, line 220: Side note - seems like we don't support translations in the App Hub?  We probably want to eventually

amcgee (commented):
Code looks good, testing now

CC @cooper-joe for the messaging

client/src/components/form/ReduxFormUtils.jsx
Birkbjo: That's correct. Something that's been discussed for a long time, but I didn't find an issue for it - will create one! It shouldn't be too bad to add, but I don't know if we want to use the same workflow as DHIS2-apps here?

amcgee:
The link isn't styled as a link which I think is undesirable.  Maybe we leave off the second line for now?  @cooper-joe 

Screen Shot 2020-09-29 at 2 37 39 PM

cooper-joe:
> CC @cooper-joe for the messaging

We should always be upfront about requirements, rather than waiting for the user to make a mistake and then telling them via an error that there were actually some hidden criteria.

So I recommend using help text underneath the input `Version number must use semantic versioning with format x.x.x (e.g. 2.3.1)`. If the validation fails this can become an error text rather than help text.

Birkbjo:
Alright! I could add some underline styling for the link though, but yeah it would be weird if it had the normal blue link color. 
In a help-text it could be styled as a link though.

cooper-joe:
> Alright! I could add some underline styling for the link though, but yeah it would be weird if it had the normal blue link color.
In a help-text it could be styled as a link though.

Oh, cool. Let's add the link to the docs for semver then 👍:
`Version number must use semantic versioning with format x.x.x (e.g. 2.3.1). <a>Read more about semantic versioning</a>.`

client/src/components/form/ReduxFormUtils.jsx
Birkbjo: Issues for that here: https://jira.dhis2.org/projects/HUB/issues/HUB-93

amcgee (approved):
There's a slight UI weirdness in the New App UI, we should address that at some point but it's not critical here

PR #380, opened 2020-10-11
fix: correctly load channels, refactor version form, fix version delete

amcgee:
This removes lots of duplication which caused inconsistent version forms.  Also, and more importantly, the loading of channels on new version creation as well as the ability to delete versions was broken and is now fixed.

There's a lot of messiness in here we should clean up, including tons of prettier and eslint errors (I fixed the ones in these files) which shouldn't be allowed to happen.

See https://community.dhis2.org/t/app-store-is-broken-cant-upload-new-app-version/40982


Birkbjo:
LGTM. 

> There's a lot of messiness in here we should clean up, including tons of prettier and eslint errors (I fixed the ones in these files) which shouldn't be allowed to happen.

Yeah. That was added after most of the client code and O haven't gotten around to clean everything up...

Birkbjo (approved):

amcgee:
@Birkbjo should we go through `next` for this or straight to `master` since it's a nasty bug in production?

Birkbjo:
Lets go through next then master. Makes it so we dont have to merge next in to master etc.

PR #385, opened 2020-10-19
feat: upgrade react 16

Birkbjo:


varl (approved):
Looks good to me, it would be nice to finally get this up to React 16 so we can start looking at the other dependencies that are lagging. :+1:

PR #400, opened 2020-11-02
chore: update node image version on Dockerfile

paschalidi:
👋  
I was playing around with some Docker security related tools the past days. One of them called [trivy](https://github.com/aquasecurity/trivy) scans your images for known vulnerabilities. 

Took the time and scanned some of our images and got a report back with some vulnerabilities on the `node:lts-slim`. Thought of taking a moment and updating the image to the `node:lts-buster-slim`. Buster is the stable 10.4 Debian release. I believe `lts-slim` uses the `stretch` which is the codename for all version 9 variations of Debian.

### How to run trivy and see the reports
1. Install trivy. You can follow the instructions from [here](https://github.com/aquasecurity/trivy#installation)
2. Scan an image, eg ` trivy node:lts-slim`. This command will return a list of known vulnerabilities on the image.

At the moment `node:lts-slim` has 
```
Total: 156 (UNKNOWN: 0, LOW: 101, MEDIUM: 13, HIGH: 30, CRITICAL: 12)
```
in comparison with the `node:lts-buster-slim` which has 
```
Total: 83 (UNKNOWN: 0, LOW: 62, MEDIUM: 5, HIGH: 16, CRITICAL: 0)
```
Keep in mind that the report is far more detailed than the information I am pasting here. 

Hope you open to contributions.


paschalidi:
@varl how can I fix the tests here? I see that some other PR are also failing with the same error.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:lts-slim as build
+FROM node:lts-buster-slim as build
```
varl, line 1: The slim image does not contain Python or other buildpack-deps, which caused a different PR to fail because Node needs Python to build `node-gyp`, so I think we want to change to the full `buster` image.

I would like us to pin the node version here. We've seen that builds fail randomly because we have used the LTS tag when the LTS version changes under our feet.

Now I have verified that everything works with node 14, so I suggest changing the line to:

`FROM node:14-buster as build`

Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN yarn workspace client pack --filename app-hub-client.tgz
 RUN tar zxvf client/app-hub-client.tgz --directory server/
 RUN mv server/package/build server/static && rm -rf server/package
 
-FROM node:lts-slim
+FROM node:lts-buster-slim
```
varl, line 16: I think it is safe to use `slim` here for the production image, but we should pin the Node version:

`FROM node:14-buster-slim`

varl (changes requested):

varl:
Thanks for taking a look at the security of our image! I've left a comment about pinning versions and switching to another image that contains the build deps for Debian, but other than that we'll happily accept this patch.

Dockerfile
paschalidi: Yep makes sense.

varl (approved):

PR #405, opened 2020-11-04
fix(api): fix apiV2 apiUrl

Birkbjo:
Seems like staging and prod are using `/api/` as the baseUrl, and as we removed both trailing and leading slashes that was treated as a relative url, so apiv2 requests resolved to `/user/api` in some cases. With the leading `/` fetch will resolve to the domain-root.


varl (approved):
Looks good to me.

PR #406, opened 2020-11-05
fix(organisationlist): replace onTouchTap with onClick

Birkbjo:


varl (approved):
Does this mean we can get rid of [this](https://github.com/dhis2/app-hub/blob/master/client/package.json#L36) too?

Birkbjo:
> Does this mean we can get rid of [this](https://github.com/dhis2/app-hub/blob/master/client/package.json#L36) too?

Yes, it's [already](https://github.com/dhis2/app-hub/blob/next/client/package.json#L34) removed.

varl:
Neat!

PR #411, opened 2020-11-09
chore(release): release organisation management

Birkbjo:


varl:
I think releasing is overdue, and we should test the staging environment and then roll out.

amcgee:
@Birkbjo how are the orgs generated?  I assume that's part of the upgrade migration?

There's a lot of weirdness there, which we should fix manually.  For instance, there is a `bluesquare` and also a `Bluesquare` organization, plus `DHIS2` and `DHIS 2` and `UiO`, plus `HISP` (no region specified), `HISP Tanzania` and `Tanzania`.  There's also `N/A`, which seems a bit odd.  Can/should we clean up these obvious confusions at the same time as releasing this feature, or immediately afterwards?

One more important bug we need to address in the above situations - organization names should be unique and CASE INSENSITIVE.  Someone shouldn't be able to create `dhis2` as an org and distribute malware under our name (or the name of any other org).  Can/should we enforce that in this change or in a subsequent (soon) change?

amcgee:
I also found a security issue we should address - anyone in an organization can invite other people to the org and also promote anyone else to be that organization's owner.  This should be restricted to only the owner, right?  Otherwise what's the point of having an owner?

amcgee:
Another issue - uploading an app allows you to specify anything as the Developer email.  This email might be something completely random, or an organization email.  That email is then silently added to the organization, even if they don't have an account which is required for adding other users, and can view / edit / delete everything to do with the organization, including other apps than the one uploaded

amcgee (changes requested):
1. Developer emails specified when uploading an app shouldn't automatically be added to the organization
2. Organization owners should be the only ones allowed to (1) add other users, (2) rename the org, and (3) promote other members to owner.  We might want to support multiple admins at some point, but that can be separate...
3. There is currently no way to (1) delete and org and (2) transfer an app from one org to another, is this intentional?

Birkbjo:
The organisations are there in production atm. The groundwork has been merged already, this "just" adds management of the organisations. 

> There's a lot of weirdness there, which we should fix manually. For instance, there is a bluesquare and also a Bluesquare organization, plus DHIS2 and DHIS 2 and UiO, plus HISP (no region specified), HISP Tanzania and Tanzania. 

Yup, and that is because organisations was just a string before, and has been transferred to be an entity. I've been thinking of going in and fixing this manually. You should not be able to create eg. `dhis2` when `DHIS2` exists, as that would result in the same slug. We could have a more sophisticated algorithm, but currently we rely on slug-uniqueness for such cases. We can also add a blocklist to prevent eg. `dhis 2`-variations.

> Developer emails specified when uploading an app shouldn't automatically be added to the organization

I agree with this, and it's been something I've questioned myself. Again not something that has been added here. But I'm open to remove it. I haven't really understood why this created a user with that email. I've also been on the fence of potentially removing this field, now that we have contact email for organisations. But it might be a use-case that users want a different email per app, but it could indeed default to the org-email.


> Organization owners should be the only ones allowed to (1) add other users, (2) rename the org, and (3) promote other members to owner. We might want to support multiple admins at some point, but that can be separate..

We've been a bit back and forth of how prohibitive we should keep the authorities of org-members. We landed on basically everyone in an organisation having the same amount of privileges. This was done both to keep it simple, and there wasn't really a use case to have more granular control, as most organisations are pretty tight-knitted - especially those who upload apps. Organisations shouldn't add arbitrary users to their organisations.
If you disagree with this, we can look into prohibiting member-rights. I agree that only the owner should be able to change the owner though. 


> There is currently no way to (1) delete and org and (2) transfer an app from one org to another, is this intentional?

Correct. The idea has been to add this later, and I probably should've created a Jira-ticket for it. The biggest hurdle has been how to handle referenced data, eg. apps - should those also be deleted?

Birkbjo:
> Developer emails specified when uploading an app shouldn't automatically be added to the organization

After investigating this a bit, it's not so straightforward to remove - or at least it will have some repercussions. As noted above, I'm unsure about the design-decision about this, but for some reason the developer-email is referenced to through an user instead of being a simple string field. So in the database the "developer" is referenced through this user-id. It also seems like this dictates the owner of the app...
Maybe @varl knows more about this? To me it would be fine to have this as a string.

Either way, removing it completely will involve some db-changes. We could easily the code that actually adds that email to the organisation - but removal of the whole developer-user-reference thing will need a bit more effort - and is more risky.

amcgee (approved):
Looks good as interface for the existing data model!

Let's get this in and open issues to make sure we follow up on the other points - I particularly don't want us to build up a bunch of data in the db that then needs to be removed or transmuted later.

varl:
> for some reason the developer-email is referenced to through an user instead of being a simple string field. So in the database the "developer" is referenced through this user-id. It also seems like this dictates the owner of the app...

Hmmmm. No, this doesn't ring any bells. Do you want the "developer" field on the app table to be a string with e-mail? If yes, then I think it references a user id to keep the data normalized. Other than that I will have to dive a bit deeper to see if I remember any discussions about this.

varl (approved):

PR #421, opened 2021-01-12
fix: improve README instructions for local installation

mediremi:
To simplify local installation, I've moved some tasks (such as creating database tables and seeding them) to `package.json` scripts and have updated the README accordingly.

I've also moved the backend configuration documentation to `.env.template`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,15 +18,14 @@ See docs in [dhis2/docker-compose/app-hub](https://github.com/dhis2/docker-compo
 ## Local Installation
 
 1. Install dependencies with `yarn install`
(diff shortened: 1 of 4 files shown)


amcgee (approved):
Looks good to me - @Birkbjo can you confirm?

Birkbjo (approved):
Nice, LGTM

amcgee:
@mediremi for reference this should not have been a `fix` - it triggered an unnecessary production deploy and version bump (this should use `docs`)

amcgee:
Also README changes probably don't need to go through `next`, but maybe worth it here since it's ahead of `master` @Birkbjo

PR #426, opened 2021-01-27
fix(imageviewer): center image when only one image

Birkbjo:
This seems to have been introduced when upgrading to react@16. Updating react-slick and the styling seems to fix it. I can add the styling as static-assets instead of using the CDN - feedback welcome.

I've also lowered the padding so that the edge of the next-image is not shown anymore. Got some feedback that this looked like weird.


mediremi (approved):

mediremi:
LGTM :ok_hand:

amcgee:
I think it would be better not to use a CDN?  Does `react-slick` not include the dependency?

amcgee:
I think it would be better not to use a CDN?  Does `react-slick` not include the dependency?

mediremi:
> I think it would be better not to use a CDN

I'd rather not use a CDN either but I guess that should be a separate PR since it would require adding a build process for the stylsheets?

Birkbjo:
I'm pretty sure webpack is configured so that we could just `import "slick-carousel/slick/slick.css";` in the entrypoint. 
Don't remember why it's using CDN for these, as they are very minimal.

amcgee:
@Birkbjo great - yeah let's do that.  Another PR is probably reasonable for that change

Birkbjo:
Cool, I've created a new PR for that here. Will merge next into this branch once that is merged to next.

PR #431, opened 2021-02-05
feat(apps): use server-side pagination and filtering 

mediremi:
Closes https://jira.dhis2.org/browse/HUB-90

Currently, app filtering is done entirely client-side and involves downloading all apps present in the system. While this is fine for the current number of apps that we have, this will not scale as the AppHub grows.

This PR implements a paginated `/api/v2/apps` endpoint that allows searching by app/organisation name as well as filtering by channel, app type and DHIS2 version. The client has also been updated to use this new endoint.

## Server side changes
- Implemented a paginated `/api/v2/apps` endpoint. Results follow the schema
    ```
    {
      apps: Array<AppModel>,
      pager: {
        page: Number,
        pageCount: Number,
        total: Number,
        pageSize: Number
      }
    }
    ```
   The pager schema follows the one already implemented by `dhis2-core`
- Added tests for `/api/v2/apps` endpoint based on the preexisting `/api/v1/apps` tests

## Client side changes
Changes to index route:
- Use new `/api/v2/apps` endpoint
- Replace use of Redux and Rxjs with declarative data query API. This marks the start of a gradual migration to data/state handling methods used by other DHIS2 apps.
- To reduce the number of network requests, channel names are hardcoded
- Pagination UI


Birkbjo:
Nice!

I've done some work on a pagination plugin as well, so we should probably decide on one solutions. See https://github.com/dhis2/app-hub/pull/384 . Kind of my fault for not pushing this further.

server/src/routes/v2/apps.js
```diff
@@ -1,24 +1,55 @@
...
+                pagination: {
+                    enabled: true,
                 },
             },
         },
-        handler: (request, h) => {
-            throw Boom.notImplemented()
+        handler: async (request, h) => {
+            let channels = ['Stable']
+            if (request.query.channels) {
```
Birkbjo, line 22: Is the idea to accept eg. `channels=Stable,Canary` here, and return the union of those channels? We haven't really how we should send "lists" to the server, but I guess delimited values are fine.

We should however verify/validate the query-objects.
Thing is we need a way to convert string into a list, this can probably be done with Joi, but it's not built in anymore. We could add something like this to our extended-Joi:

```
Joi.extend({
  base: joi.array(),
  name: 'stringArray',
  coerce: (value, state, options) => (value.split ? value.split(',') : value)
})
```

Then we could use that with the custom Joi.filter() and Filter-plugin.

```
  validate: {
                query: Joi.object({
                    channels: Joi.filter(Joi.stringArray().items(Joi.valid('Stable', 'Canary'))).description(
                        'Filter by channel'
                    ),
                    types: Joi.filter(Joi.stringArray().items(Joi.valid(...APPTYPES))).description(
                        'Filter by app types'
                    ),
```

In handler:

```
const channels = request.plugins.queryFilter.getFilter('channels').value
```


The idea when implementing this filter-mechanism was that it could be used transparently" by eg doing `?channels=in:Stable,Canary.` Then you would get the filters applied to your query by doing `filters.applyToQuery(query)`. However, that is not implemented for lists yet (works for simple values though). This query might also be a bit too complicated for that anyway, at least for now.

Anyways, that might be a bit out of scope, but it would be great to get some validation on the query-parameters.

client/src/components/apps/Apps.jsx
```diff
@@ -0,0 +1,76 @@
...
+    Stable: 'Stable',
+    Development: 'Development',
+    Canary: 'Canary',
+}
+
+const types = {
+    APP: 'Standard app',
+    DASHBOARD_WIDGET: 'Dashboard app',
+    TRACKER_DASHBOARD_WIDGET: 'Tracker widget',
+}
```
Birkbjo, line 19: Maybe we should move this to `apiConstants.js`?

client/src/components/apps/Apps.jsx
```diff
@@ -0,0 +1,76 @@
...
+import Filters from './Filters'
+import Pagination from './Pagination'
+
+const channels = {
+    Stable: 'Stable',
+    Development: 'Development',
+    Canary: 'Canary',
+}
+
+const types = {
```
Birkbjo, line 15: These are already defined in `default.config.js`, under `appTypeToDisplayName`. Might not be the most intuitive place to have it, but we should keep it to one place. Maybe that should also be moved to apiConstants?

server/src/routes/v2/apps.js
mediremi: > Is the idea to accept eg. channels=Stable,Canary here, and return the union of those channels?
Yep

> Anyways, that might be a bit out of scope, but it would be great to get some validation on the query-parameters.
I'll implement the query validation now and we can defer the `applyToQuery` changes for another PR

client/src/components/apps/Apps.jsx
mediremi: Yep let's move both constants to `apiConstants`

server/src/routes/v2/apps.js
```diff
@@ -1,24 +1,76 @@
...
+const { getApps } = require('../../data')
+const { convertAppsToApiV1Format } = require('../v1/apps/formatting')
+const { filterAppsBySpecificDhis2Version } = require('../../utils/filters')
+const CustomJoi = require('../../utils/CustomJoi')
 
-const AppModel = require('../../models/v2/out/App')
+const CHANNELS = ['Stable', 'Development', 'Canary']
+const APPTYPES = ['APP', 'DASHBOARD_WIDGET', 'TRACKER_DASHBOARD_WIDGET']
+
+const StringArrayJoi = Joi.extend({
```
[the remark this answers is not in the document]
Birkbjo: I'm not sure if we want paging like this and not just "load more"?

client/src/components/apps/Pagination.jsx
```diff
@@ -0,0 +1,33 @@
...
+        margin: '0 0.5em',
+    },
+}
+
+const Pagination = ({ page, pageCount, onPageChange }) => (
+    <>
+        <FlatButton
+            label="Previous page"
+            onClick={() => onPageChange(page - 1)}
+            disabled={page == 1}
```
Birkbjo, line 16: Hate to be nit-picky, but generally prefer '==='

server/src/routes/v2/apps.js
```diff
@@ -1,24 +1,76 @@
...
+                },
+                h.context.db
+            )
+            const filteredApps = filterAppsBySpecificDhis2Version(
+                apps,
+                request.query.dhis_version
+            )
+            const pager = request.plugins.pagination
+            const result = convertAppsToApiV1Format(filteredApps, request)
+            return h.paginate(pager, {
```
Birkbjo, line 70: I see that we could simplify this api by removing the pager-param. The toolkit (h) should have access to the request so we should be able to get the pager without having to pass it. Could potentially have "pager" as an optional override in the options object. Probably not something we should change here though.

server/src/routes/v2/apps.js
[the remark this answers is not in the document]
mediremi: > I'm not sure if we want paging like this and not just "load more"? I think we can also afford to load more than 12 apps, but I agree 12 makes sense with this approach.

Do you mean something like infinite scroll where apps would be loaded automatically as soon as the user reaches the bottom of the page? My only concern with something like that is that it can 'break' the back button by not restoring the user to their previous scroll position - though tbf I've just realised that since we're not storing pagination details in query params the current state of this PR breaks the back button too. What are your thoughts on using something like [`use-query-params`](https://github.com/pbeshai/use-query-params) to store pagination and filtering state in the URL like how I've done in https://github.com/dhis2/sms-configuration-app/pull/217? 

I picked 12 as [we render 3 app cards per row on smaller screens and 4 otherwise](https://github.com/dhis2/app-hub/blob/b4496df573e8587678e2f7e91f90049bfef5425d/client/src/components/appCards/AppCards.jsx#L131) so the next number we can bump the page size up to is 24. With the UI redesign having a two apps per row layout we will have a wider range of `pageSize` to choose from.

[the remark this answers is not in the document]
Birkbjo: Infinite scroll can indeed be annoying when the scroll position is reset, and I guess it would be more complicated to implement - and as we are looking to redesign in in the not too distant future, I think this approach is fine.

Yup I thought about that as well, would be great if we could use the query-params.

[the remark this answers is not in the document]
mediremi: I've updated the filtering state managent to use query params and I've fixed the subheader back button on the app view page.

Birkbjo (approved):
Good work, really nice to get the filter in the query-params!

mediremi:
@varl @amcgee Any thoughts/feedback? If not I'll merge later today.

server/src/data/getApps.js
```diff
@@ -5,25 +5,51 @@ const debug = require('debug')('apphub:server:data:getApps')
...
+            }
+
+            if (types.length > 0) {
+                builder.where(builder => {
+                    builder.where('type', types[0])
+                    types.slice(1).forEach(type => {
+                        builder.orWhere('type', type)
+                    })
+                })
+            }
```
amcgee, line 43: These should be case-insensitive, are they? - I haven't tested, but it appears that `?channels=stable` won't work (because it expects `Stable`)

client/default.config.js
```diff
@@ -31,5 +31,11 @@ module.exports = {
             DASHBOARD_WIDGET: 'Dashboard',
             TRACKER_DASHBOARD_WIDGET: 'Tracker Dashboard',
         },
+        appChannelToDisplayName: {
+            Stable: 'Stable',
+            Development: 'Development',
+            Canary: 'Canary',
+        },
```
amcgee, line 38: I think the canonical names should be lowercase, it's weird to have a proper-cased enum.

server/src/routes/v1/apps/handlers/getAllApprovedApps.js
```diff
@@ -30,22 +30,22 @@ module.exports = {
...
 
-        //default to Stable if not specified. Or undefined if we want to fetch all channels
-        const channel =
+        //default to Stable if not specified. Or empty list if we want to fetch all channels
+        const channels =
             request.query.channel === 'All'
-                ? undefined
-                : request.query.channel || 'Stable'
+                ? []
+                : [request.query.channel || 'Stable']
```
amcgee, line 37: Here is where we're being overly case-sensitive, I think

server/src/routes/v2/apps.js
```diff
@@ -1,24 +1,67 @@
...
+const AppModel = require('../../models/v1/out/App')
+const { AppStatus } = require('../../enums')
+const { getApps } = require('../../data')
+const { convertAppsToApiV1Format } = require('../v1/apps/formatting')
+const { filterAppsBySpecificDhis2Version } = require('../../utils/filters')
+const Joi = require('../../utils/CustomJoi')
 
-const AppModel = require('../../models/v2/out/App')
+const CHANNELS = ['Stable', 'Development', 'Canary']
+const APPTYPES = ['APP', 'DASHBOARD_WIDGET', 'TRACKER_DASHBOARD_WIDGET']
```
amcgee, line 9: Similarly for app types, but capitalized snake casing is a little more normalized - in any case we should be consistent

amcgee (approved):
I scanned the code quickly and it looks good to me, nice change!!  I didn't look into any details.  I found one API issue (I think) which it would be nice to address.

server/src/data/getApps.js
mediremi: `Stable` is how the channel name is stored in the DB currently. I'll make a migration to rename it to `stable` :+1:

server/src/routes/v2/apps.js
mediremi: I think the app types are named that way to reflect [how `dhis2-core` stores them](https://github.com/dhis2/dhis2-core/blob/ea76fa86409613c9766d4508d65c88ac55c413c3/dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/AppType.java). Is there a risk that renaming them would impact core (through its use of v1 APIs)? I guess we could convert e.g. `dashboardWidget` to `DASHBOARD_WIDGET` in v1 responses

Birkbjo: I think we should keep these consistant with the dhis2-core naming.

amcgee: Sounds good to me, better to be safe and consistent

server/src/data/getApps.js
amcgee: OK 👍 - if you want to do that in a separate change that could be OK too

mediremi: [This commit](https://github.com/dhis2/app-hub/pull/431/commits/f9b90f543276d27dcbe8c14bbef44f706c98f386) has the changes for renaming channels.

Should I also guarantee case-insensitivity in the API by doing something like this?
```js
builder.where(builder => {
    builder.where('channel_name', channels[0].toLowerCase())
    channels.slice(1).forEach(channel => {
        builder.orWhere('channel_name', channel.toLowerCase())
    })
})
```

client/default.config.js
varl: Agreed.

server/src/data/getApps.js
Birkbjo: > Should I also guarantee case-insensitivity in the API by doing something like this? 

I vote yes!

mediremi: Done :heavy_check_mark:

PR #437, opened 2021-02-12
fix(createApiUser): fix server crash when audience is changed

Birkbjo:
For some reason the uuid of the `user_external_id` was hardcoded when adding apiUser, so if the audience was changed it would fail with unique-constraint error, this would also crash the server on startup. The transaction-promise was not returned, so the chained .catch was never reached. This allows multiple audiences for the api-user.


mediremi (approved):

Birkbjo:
Updated branch to latest next

server/src/server/create-api-user.js
```diff
@@ -27,7 +29,7 @@ exports.createApiUser = async knex => {
     if (externalUserId.length === 0) {
         await knex('user_external_id').insert([
             {
-                id: '46e00c3b-4668-4b93-bc6f-a46bc21f1e5d',
```
varl, line 30: As a note, this same UUID is used in: `server/seeds/01_organisation-developers.js`

I don't know about the history here, but those seeds look hand-rolled which means it was added manually in both places by a developer. If it is because of a valid reason or not I don't know.

To me it seems like a copypaste, and I agree that it should be a generated `uuid`.

varl (commented):
LGTM!

PR #440, opened 2021-02-15
fix(userValidation): throw specific error if user exists in DB but no id

Birkbjo:
### Background

While testing the new auth-flow @amcgee experienced a 500-error. The reason for the error was that the `external_id` (`decoded.sub` in the token) was not found in the DB, but the email was. This was introduced with the new auth-flow as we do not have access to the email from the token - and thus we rely upon the external_id to be in the database. For some unknown reason the staging-DB does not have all external IDs. 

## Solution
This PR does not really solve the issue. You would think that this could've easily been solved by adding `external_id` to the user with the matching email in the DB, but we should not do that as that would be the same as automatic linking accounts.

Therefore I've just added a more specific error so the user might be able to contact us or try another provider. Note that I've removed the try-catch in create-user as I'm generally against such "error-rewrites", and it's needed for `wrapError` to map it to the correct error-class. I've also changed `conflict`-errors thrown from the validation-function to `unauthorised` as that should help indicate that the error is with the auth and not the actual payload/request.

If this happens to a user in the wild  we need to manually add the external_id in the DB for the user it affects. However, I've checked the prod-DB and almost all users have a linked external ID.

## Complications and future fixes

This should be a pretty rare occurrence, and I don't know why the staging-DB is so incomplete in regards to missing external_ids. However there's one problem where this might occur if this is pushed to the live server.
Currently we are creating new users through the developer-email when uploading an app. This results in different problems and has been discussed before. See [here](https://github.com/dhis2/app-hub/pull/411#pullrequestreview-530251570) and my answer [here](https://github.com/dhis2/app-hub/pull/411#issuecomment-728180099)).

This error would happen to all users that try to log in with an email that has been added like this.


server/src/services/apiKey.js
```diff
@@ -0,0 +1,23 @@
+const { NotFoundError } = require('../utils/errors')
```
mediremi, line 1: Looks like this file was for a different PR

Birkbjo: Woops...

mediremi (approved):

server/src/security/createUserValidationFunc.js
```diff
@@ -47,22 +47,31 @@ const createUserValidationFunc = (db, audience, auth0ManagementClient) => {
...
+                                    email: userInfo.email,
+                                    name: userInfo.name,
+                                },
+                                trx
+                            )
+                        } catch (e) {
+                            // if user exist and not current external-id throw a specific error
+                            Bounce.ignore(wrapError(e), UniqueViolationError)
+                            throw Boom.unauthorized(
+                                'Email is already registered with another provider.'
```
amcgee, line 72: Should maybe add "please contact DHIS2 at <xyz> if you think this is a mistake."

PR #445, opened 2021-03-03
fix(getApps): don't call toLowerCase on app type

mediremi:


Birkbjo:
Did we not want case insensitive channels? Or is it lower-cased somewhere else?

mediremi:
We do want case-sensitive *channels* (as handled [here](https://github.com/dhis2/app-hub/blob/09181493095402e4def5d9c207d94583789f2719/server/src/data/getApps.js#L27-L34)), the issue is that *app types* are not case insensitive since we use the same names as from `dhis2-core` (which are in `THIS_STYLE`).

I forgot that we store app types differently in the DB when I made changes to `getApps` for ignoring channel names casing and hence right now [staging](https://staging.apps.dhis2.org/) does not load any apps.

Birkbjo:
> We do want case-sensitive _channels_ (as handled [here](https://github.com/dhis2/app-hub/blob/09181493095402e4def5d9c207d94583789f2719/server/src/data/getApps.js#L27-L34)), the issue is that _app types_ are not case insensitive since we use the same names as from `dhis2-core` (which are in `THIS_STYLE`).
> 
> I forgot that we store app types differently in the DB when I made changes to `getApps` for ignoring channel names casing and hence right now [staging](https://staging.apps.dhis2.org/) does not load any apps.

Ah of course, this is type and not channel!

Birkbjo (approved):

PR #447, opened 2021-03-03
feat: apiKey generation

Birkbjo:
This implements an API to generate API-keys. The generated API-key can be used instead of an OAuth token to identify the request from the user, provided in the `x-api-key`-header in the request. 

### Implementation 
See https://jira.dhis2.org/projects/HUB/issues/HUB-100 for more information about the considered implementation options. I decided to generate the key using uuid-v4 and hashing this with sha-256. This was the simplest solution; we already have uuid-v4 as a dependency, it's cryptographically secure (backed by [crypto.randomFill](https://nodejs.org/api/crypto.html#crypto_crypto_randomfill_buffer_offset_size_callback)), it's fast and we do not need secrets. However it means that the user cannot see the API-key after first generation. However it's simple to revoke and generate a new one.

Note that this PR does not enable API-keys for all other endpoints. To do that we simply need to add this to the to the route-config. Might look into generating a default for this.

```javascript
 auth: {
    strategies: ['token', 'api-key'],
},
``` 

I've implemented the data as simple as possible, so currently the db table only consists of (api_key, user_id), and the user_id is unique. However, we could potentially support multiple API-keys per user, and have some kind of name or description attached to the key. In that case we should probably have some kind of uuid in that table as well. If this is something we may want down the line it's easier to implement right away. So is this something we should add now?

### Caveats

As mentioned in a comment in the Jira-issue: this does not support roles. That means that a user cannot use the `manager`-permissions to edit other apps other than in their organisation.

### DIFF
--- a/server/migrations/20210209171106_api-key.js
+++ b/server/migrations/20210209171106_api-key.js
@@ -1,6 +1,6 @@
 exports.up = async function(knex) {
     await knex.schema.createTable('user_api_key', table => {
-        table.string('api_key', 64).primary() // sha256 in hex representation is 64 chars
+        table.string('hashed_api_key', 64).primary() // sha256 in hex representation is 64 chars
 
         table
             .uuid('user_id')
--- a/server/migrations/20210209171106_api-key.js
+++ b/server/migrations/20210209171106_api-key.js
@@ -1,0 +1,24 @@
+exports.up = async function(knex) {
+    await knex.schema.createTable('user_api_key', table => {
+        table.string('api_key', 64).primary() // sha256 in hex representation is 64 chars
+
+        table
+            .uuid('user_id')
+            .notNullable()
+            .unique()
+
+        table
+            .foreign('user_id')
+            .references('id')
+            .inTable('users')
+
+        table
+            .timestamp('created_at', true)
+            .defaultTo(knex.fn.now())
+            .notNullable()
+    })
+}
+
+exports.down = async function(knex) {
+    await knex.schema.dropTable('user_api_key')
+}
--- a/server/src/security/createApiKeyValidationFunc.js
+++ b/server/src/security/createApiKeyValidationFunc.js
@@ -1,0 +1,43 @@
+//const { find: findAPIKey } = require('../services/apiKey')
+const Boom = require('@hapi/boom')
+const Bounce = require('@hapi/bounce')
+const debug = require('debug')('apphub:server:security:apiKeyValidation')
+const crypto = require('crypto')
+
+const createApiKeyValidationFunc = db => {
+    return async (apiKey, request, h) => {
+        debug('Validate api-key')
+        const credentials = {}
+        try {
+            const hashedKey = crypto
+                .createHash('sha256')
+                .update(apiKey)
+                .digest('hex')
+
+            const user = await db('users')
+                .select('users.*')
+                .innerJoin('user_api_key', 'user_api_key.user_id', 'users.id')
+                .where({
+                    api_key: hashedKey,
+                })
+                .first()
+
+            if (!user) {
+                throw Boom.unauthorized()
+            }
+            credentials.email = user.email
+            credentials.userId = user.id
+            credentials.roles = []
+
+            return {
+                isValid: true,
+                credentials,
+            }
+        } catch (e) {
+            Bounce.rethrow(e, 'system')
+            return { isValid: false }
+        }
+    }
+}
+
+module.exports = createApiKeyValidationFunc
--- a/server/src/security/createApiKeyValidationFunc.js
+++ b/server/src/security/createApiKeyValidationFunc.js
@@ -41,6 +41,7 @@ const createApiKeyValidationFunc = db => {
             }
         } catch (e) {
             Bounce.rethrow(e, 'system')
+            debug(e)
             return { isValid: false }
         }
     }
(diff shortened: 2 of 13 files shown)


mediremi:
> I've implemented the data as simple as possible, so currently the db table only consists of (api_key, user_id), and the user_id is unique. However, we could potentially support multiple API-keys per user, and have some kind of name or description attached to the key. In that case we should probably have some kind of uuid in that table as well. Is this something we want now?

Could we drop the unique `user_id` constraint and make the `api_key` the primary key?

Birkbjo:
> > I've implemented the data as simple as possible, so currently the db table only consists of (api_key, user_id), and the user_id is unique. However, we could potentially support multiple API-keys per user, and have some kind of name or description attached to the key. In that case we should probably have some kind of uuid in that table as well. Is this something we want now?
> 
> Could we drop the unique `user_id` constraint and make the `api_key` the primary key?

It already is the primary key. However if we want multiple api-key per user I think it would be easier to handle if it had a unique-key that is not the API-key itself. Or, when I think about it I guess we could safely use the hashed-key it self? My train of thought was that it would be easier to have an identifier that is not the API-key itself.

mediremi:
Ah sorry I forgot that the key was hashed. What I meant was a table schema where the primary key is what we look up for validating the key i.e. the hased version of the key. That way all the middleware has to do is `userId = SELECT user_id FROM user_api_key WHERE hashed_key = ?` and as the primary key is indexed this should be pretty fast.

server/src/security/createApiKeyValidationFunc.js
```diff
@@ -0,0 +1,48 @@
+//const { find: findAPIKey } = require('../services/apiKey')
```
mediremi, line 1: :eyes:

Birkbjo:
> Ah sorry I forgot that the key was hashed. What I meant was a table schema where the primary key is what we look up for validating the key i.e. the hased version of the key. That way all the middleware has to do is `userId = SELECT user_id FROM user_api_key WHERE hashed_key = ?` and as the primary key is indexed this should be pretty fast.

That's kind of how it works atm. https://github.com/dhis2/app-hub/pull/447/files#diff-25cd0e7206ed60e26c252745147d2e718d2ad9e881c0be0894f1e7555e8f8f2eR42 . However we need the email as well, so a separate query is ran as well. At first all this was in one query, but I did not want the API-service to include User-related tables, as we eg. want to remove email from that pretty soon regardless. 

What I mean with the separate uuid is if we want to be able to reference an API-key without exposing the key it self. However if we keep it unique per user that is not needed, as user_id is enough.

server/src/security/createApiKeyValidationFunc.js
Birkbjo: Put this out of draft too soon!

mediremi:
> What I mean with the separate uuid is if we want to be able to reference an API-key without exposing the key it self. However if we keep it unique per user that is not needed, as user_id is enough.

So the concern is exposing `sha256(api key)` in - for example - a `/user/api-keys` endpoint that would be used by the client to allow the user to view/delete keys? That should be fine since SHA256 is a [cryptographic hash function](https://en.wikipedia.org/wiki/Cryptographic_hash_function) and so has pre-image resistance, which means given a hash `h` it's practically impossible to find `m` such that `h = H(m)`.

Birkbjo:
> > What I mean with the separate uuid is if we want to be able to reference an API-key without exposing the key it self. However if we keep it unique per user that is not needed, as user_id is enough.
> 
> So the concern is exposing `sha256(api key)` in - for example - a `/user/api-keys` endpoint that would be used by the client to allow the user to view/delete keys? That should be fine since SHA256 is a [cryptographic hash function](https://en.wikipedia.org/wiki/Cryptographic_hash_function) ...

Yup, I know, it is. It's not that I'm too concerned exposing it, however that it's sha-256 is an implementation detail. It's also more as a design decision - to have another identifier other than the key.

mediremi:
I agree that from a design point of view we do need a UUID on the keys. We can make the UUID the primary key and then use `CREATE INDEX` on the `api_key` column to make lookup by hashed API key as fast as if it were the primary key.

server/migrations/20210209171106_api-key.js
```diff
@@ -0,0 +1,24 @@
+exports.up = async function(knex) {
+    await knex.schema.createTable('user_api_key', table => {
+        table.string('api_key', 64).primary() // sha256 in hex representation is 64 chars
```
mediremi, line 3: Should we call this column `sha256_api_key` to make it clearer that it's not the actual API key?

Birkbjo: Sure, or just hashed_api_key?

Birkbjo:
> I agree that from a design point of view we do need a UUID on the keys. We can make the UUID the primary key and then use `CREATE INDEX` on the `api_key` column to make lookup by hashed API key as fast as if it were the primary key.

Well, we would still have an `unique`-constraint on the column, which generates an index already.

server/migrations/20210209171106_api-key.js
mediremi: Even better :+1:

mediremi:
> Well, we would probably still have an unique-constraint on the key-column, which generates an index already.

Yep you're right I forgot that for unique to work it needs to create an index.

mediremi (approved):

varl (approved):
:+1:

amcgee:
I'm late to the party but this looks awesome, nice work!  🎉

PR #450, opened 2021-03-09
fix(version): fix app version upload

Birkbjo:
Adds support for uploading app-version with api-key. 

There was a few bugs when testing this due to not actually using the transaction in the helper-function to create the version. This resulted in partially DB-creations. Eg. if you tried to upload with an invalid channel, the version was created in the `app_version`-table, but it failed later when trying to add the version to the channel-table. This resulted in a 500-error if you tried to upload the version again (with fixed channel), even though the version would not show up in the API. 

I intend to merge this using regular merge as the commits are standalone. They are related but does not depend on each other.

### DIFF
--- a/server/src/routes/v1/apps/handlers/createAppVersion.js
+++ b/server/src/routes/v1/apps/handlers/createAppVersion.js
@@ -26,7 +26,9 @@ module.exports = {
     method: 'POST',
     path: '/v1/apps/{appId}/versions',
     config: {
-        auth: 'token',
+        auth: {
+            strategies: ['token', 'api-key'],
+        },
         tags: ['api', 'v1'],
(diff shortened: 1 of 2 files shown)


server/src/routes/v1/apps/handlers/createAppVersion.js
```diff
@@ -118,87 +120,80 @@ module.exports = {
 
         let versionId = null
```
mediremi, line 121: I think this can be removed and its sole use can be replaced with `appVersion.id`

mediremi (approved):

mediremi:
LGTM :tada: - I suggested a very minor change that can be delayed to a later refactor

PR #454, opened 2021-03-17
feat(apikey): frontend for api key generation

Birkbjo:


client/src/components/user/ApiKey.jsx
```diff
@@ -0,0 +1,172 @@
...
+    if (error) {
+        return 'Failed to load API-key status'
+    }
+    if (!data) {
+        return <Spinner />
+    }
+
+    return data.hasApiKey ? (
+        <ApiKeyDisplay
+            createdAt={data.createdAt}
```
mediremi, line 159: `data.createdAt` is inserted by the client in `handleGenerateKey` and is not present in the server response. This means that if users refresh the page, they see `API key was generated at Invalid Date`.

My suggestion is to add a `created_at` column to the `user_api_key` table and to change `getApiKeyByUserId` in `services/apiKey.js` so that it returns `{ hash: "...", createdAt: "..." }`.

client/src/components/user/ApiKey.jsx
```diff
@@ -0,0 +1,172 @@
...
+            </Card>
+        </div>
+    )
+}
+
+const GenerateApiKey = props => {
+    return (
+        <div>
+            <NoteBlock>
+                Once an API-key is generated, you will not be able to view it
```
mediremi, line 48: To emphasise the 'treat API keys as passwords' part, I think we can remove the first sentence (users see it after generating an API key anyway).

Also a more warning-like colour (orange/red) would make this stand out more:
image

mediremi (approved):
LGTM :tada:

PR #456, opened 2021-03-24
fix: show query in input when returning from search result

mediremi:
Filtering apps using a search query, clicking on an app and then returning to the previous page results in a blank searchbox even though the URL query params contains the correct query.

This PR fixes this issue by passing the query value on first render to the search filter.


Birkbjo (approved):
LGTM

PR #458, opened 2021-04-01
feat: UI redesign

mediremi:
Closes https://jira.dhis2.org/browse/HUB-107 and https://jira.dhis2.org/browse/HUB-33

Demo: https://redesign.eu-west-1.elasticbeanstalk.com/

## Frontend changes
- Replaced all `material-ui` components with ones from `@dhis2/ui`
- Separate [page components](https://github.com/dhis2/app-hub/tree/752e4a273b5e6443daf39d51116b4954e94f3e48/client/src/pages) from [reusable shared components](https://github.com/dhis2/app-hub/tree/752e4a273b5e6443daf39d51116b4954e94f3e48/client/src/components)
- Removed DHIS2 logo from `assets` folder and instead load from `@dhis2/ui`
- [Used shim for `@dhis2/app-runtime` in order not to have to add as dependency](https://github.com/dhis2/app-hub/blob/752e4a273b5e6443daf39d51116b4954e94f3e48/client/webpack.config.js#L78-L81)
    + Introducing additional shim for `@dhis2/d2-i18n` would have reduced bundle size by 40KB but would have been more complex due to need for shimming `i18n.addResource`
- Updated `@dhis2/cli-style`
- Created custom [dropdown button](https://github.com/dhis2/app-hub/blob/752e4a273b5e6443daf39d51116b4954e94f3e48/client/src/components/Header/DropdownButton/DropdownButton.js) and [flyout menu](https://github.com/dhis2/app-hub/blob/752e4a273b5e6443daf39d51116b4954e94f3e48/client/src/components/Header/DropdownMenu/DropdownMenu.js) components
    + Using [existing `@dhis2/ui` components](https://ui.dhis2.nu/demo/?path=/story/actions-buttons-dropdown-button--with-menu) was not suitable due to lack of keyboard accessibility
    + As part of future PR, need to handle `ESC` key for closing menu
- In order to create keyboard accessible links that look like `@dhis2/ui` buttons, I used `<a tabindex="-1"><Button>Label</Button</a>`. This way only the inner button is keyboard focusable, but when pressing enter the browser still triggers `onClick` of `<a>`
- [Implemented an `AlertsProvider`](https://github.com/dhis2/app-hub/blob/752e4a273b5e6443daf39d51116b4954e94f3e48/client/src/components/AlertsProvider/AlertsProvider.js) similar to the one used by the app platform
   + To make consumption easier, [three hooks were created: `useAlert`, `useSuccessAlert`, `useErrorAlert`](https://github.com/dhis2/app-hub/blob/752e4a273b5e6443daf39d51116b4954e94f3e48/client/src/lib/use-alert.js)
   + Added 'undo' action to alerts when action is reversible
      - See https://alistapart.com/article/neveruseawarning/
      - For non-reversible actions used `window.confirm` but ideally should either i) use modal from `@dhis2/ui` or (even better imo) ii) change 'delete' backend actions to reversible 'archive' actions (i.e. no more permanent deletion)
- Added a `client/src/lib` folder where non-React code (e.g. hooks, form validators and other small reusable functions) lives
- Renamed `.jsx` files to `.js`
- Updated React to version 17 and updated babel jsx transformer
    + [No longer need to `import React` at the top of every jsx file](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)
- Added `useQueryV1` hook for v1 API calls
- Material icons icon font replaced with individually imported SVGs in order to reduce bundle size by ~170KB
- Bundle size reduced from 1.5MB to 765KB :tada: 

:warning: **Potentially controversial changes that I am happy to revert:**
- [I used webpack's resolve alias](https://github.com/dhis2/app-hub/blob/752e4a273b5e6443daf39d51116b4954e94f3e48/client/webpack.config.js#L75-L77) to treat `client/src`, `client/config` and `client/assets` as top level packages – allowing us to simplify complex relative imports such as `import api from '../../../api'` into `import api from 'src/api'`
    + The main downside to this approach is that testing using nodejs is more complex, though this can be mitigated by moving the resolve aliases from webpack to [babel](https://github.com/tleunen/babel-plugin-module-resolver)
    + It may also make the code more confusing to outside developers due to `src`, `config` and `assets` being imported as if they were npm packages
- Changed behaviour of [v2](https://github.com/dhis2/app-hub/blob/752e4a273b5e6443daf39d51116b4954e94f3e48/client/src/api/AppHubAPI.js#L59-L62) and [v1](https://github.com/dhis2/app-hub/blob/752e4a273b5e6443daf39d51116b4954e94f3e48/client/src/api/index.js#L145-L147) API clients by throwing `new Error(response.statusText)` instead of `response` in order to make error rendering with `NoticeBox` and alerts easier (just needed to use `error.message` as content)

## Backend changes
- [Verify that the protocol of a source code URL is `http:` or `https:`](https://github.com/dhis2/app-hub/blob/752e4a273b5e6443daf39d51116b4954e94f3e48/server/src/data/updateApp.js#L31-L38)
   + To guard against XSS achieved using `javascript:` links
- [Implemented `v2` create app handler closely based on `v1` version](https://github.com/dhis2/app-hub/blob/7ebc2d5b2b4e9edb390e01a920b942a99a7f2f55/server/src/routes/v2/apps.js#L78-L208) (see tests [here](https://github.com/dhis2/app-hub/blob/7ebc2d5b2b4e9edb390e01a920b942a99a7f2f55/server/test/routes/v2/apps.js))
   + Main difference is that an existing organisation ID must be passed (instead of an organisation name, which was used to create non-existent organisations)
   + Also does not implicitly create app dev user (which occurred if developer email address did not correspond to existing user)
   + Enforces logo and app description

## Future work
+ Extract shared styles into own CSS files and then use CSS modules `composes` feature 
  - `src/shared-styles` folder?
+ Remove redux, react-redux, redux-logger, rxjs and redux-observable
  - Only used for loading user
  - Would shave ~50KB off bundle size
  - Could have user context instead?
+ Fix all formatting issues for server files and add linting to GitHub workflow and git hooks
+ Figure out why validation error messages are not rendered for `SingleSelectFieldFF` (as used [here](https://github.com/dhis2/app-hub/blob/3b5cb71426150cfef3322fa296e8d2a21b7224ab/client/src/pages/UserAppVersionEdit/UserAppVersionEdit.js#L141-L152))
+ Implement API to replace existing logo instead of having to do it in 3 steps
  (delete existing, upload new image, set new image as logo)
  - https://github.com/dhis2/app-hub/blob/3b5cb71426150cfef3322fa296e8d2a21b7224ab/client/src/pages/UserApp/DetailsCard/DetailsCard.js#L28-L37


server/src/routes/v2/apps.js
```diff
@@ -64,4 +87,207 @@ module.exports = [
...
+            validate: {
+                payload: CreateAppModel.payloadSchema,
+            },
+            plugins: {
+                'hapi-swagger': {
+                    payloadType: 'form',
+                },
+            },
+        },
+        handler: async (request, h) => {
```
Birkbjo, line 112: I think we should extract some of these to an Apps-Service, and structure it more like organisation routes. Handler is still too complex imo

server/src/routes/v2/apps.js
```diff
@@ -64,4 +87,207 @@ module.exports = [
...
+                isManager = currentUserIsManager(request)
+            } catch (err) {
+                throw Boom.unauthorized('No user found for the request')
+            }
+
+            let appId = null
+            let versionId = null
+            let iconId = null
+
+            const trx = await knex.transaction()
```
Birkbjo, line 150: Likewise, I think we should move this logic to compose app-services? And use `knex.transaction(trx => )` API, rather than committing and aborting manually.

Birkbjo:
> Implement API to replace existing logo instead of having to do it in 3 steps
(delete existing, upload new image, set new image as logo)

You should be able to do this the current way? Eg. set `logo: true` in a PUT-request to `images/id`

Birkbjo:
Good job, this looks very nice! It's a huge PR so will take some more time to go through it more thoroughly. I do think that the organisation drop-down etc. is a bit hidden away. This is also a question for @cooper-joe , but I think a left-side menu on the "your apps" would make sense and be easier for users.

mediremi:
> You should be able to do this the current way? Eg. set logo: true in a PUT-request to images/id

That works for existing images (and is what `await api.updateImage(appId, newLogoId, { logo: true })` does) but when a user uploads a new logo through the frontend, this is represented on the backend through 3 actions:

1. If a current logo exists, delete it
2. Upload new logo as standard image
3. Set this image to be a logo

I think that it would be better if we had one atomic API call that did all three as if any of the steps fails the app state will not be consistent with what the user expected.

mediremi:
> Good job, this looks very nice!

Thanks! :smile: 

> I do think that the organisation drop-down etc. is a bit hidden away.

Yep I agree - which is why I tried to put the more commonly used pages in the headerbar and then relegated the more obscure features (organisations and API keys) to a GitHub-inspired dropdown.

I guess I could copy the sidebar design from the app management design specs?

image

cooper-joe:
> This is also a question for @cooper-joe , but I think a left-side menu on the "your apps" would make sense and be easier for users.

> I guess I could copy the sidebar design from the app management design specs?

I think that would work, yeah, though I'm not sure it's necessary. I was thinking that "Your organisations" was something that would only be accessed a handful of times – at the time of account creation and maybe some updates after that. 

If there are a few other useful things to keep in that left side menu then I don't see an issue with also having 'Your organisations' in there too though.

server/src/routes/v2/apps.js
```diff
@@ -60,8 +70,140 @@ module.exports = [
...
+                        maxDhisVersion: maxDhisVersion || '',
+                        channel,
+                        appName: name,
+                        description: description || '',
+                    },
+                    trx
+                )
+
+                const { logo } = payload
+                const logoMetadata = logo.hapi
```
Birkbjo, line 179: Lets integrate the newly merged mimeType-validation : https://github.com/dhis2/app-hub/pull/461
Should be just a matter of 
`validateImageMetadata(request.server.mime, logoMetadata`

server/src/models/v2/in/CreateAppModel.js
```diff
@@ -0,0 +1,38 @@
...
+const { isSemver } = require('../../helpers')
+
+const { AppTypes } = require('../../../enums')
+
+const CreateModelAppData = Joi.object().keys({
+    name: Joi.string(),
+    description: Joi.string(),
+    appType: Joi.string().valid(...AppTypes),
+    sourceUrl: Joi.string().uri(),
+    developer: Joi.object().keys({
```
Birkbjo, line 11: I think this model will change a bit. Depending on what we decide for https://github.com/dhis2/notes/discussions/260 , developer-information will probably change. Do we merge this before then, or decide on an API for before?

I think we should remove the `developer`-object and keep `developerEmail` and `organisationId` outside of the object.

server/src/routes/v2/apps.js
```diff
@@ -19,27 +28,28 @@ module.exports = [
...
+                    enabled: true,
                 },
                 pagination: {
                     enabled: true,
                 },
             },
         },
         handler: async (request, h) => {
-            const channels = request.plugins.queryFilter.getFilter('channels').value
+            const channels = request.plugins.queryFilter.getFilter('channels')
```
Birkbjo, line 51: Overall it seems like this is implemented like the old v1 API-structure, which we wanted to move away from. There's some abstractions that are used for the Organisation that I would like to see be used more, and see how it performs in a more complex case like this. It may be that the abstractions are not fit for it, as Organiastion are very simple compared to this. But I would like to see how it worked out. 

Maybe we should try to do a refactor after, as the primary focus for this PR is not the `v2`-API. I do think that the main changes you've done could be done directly in the `v1` API, as that needs fixing for the user-problem either way. I also don't think we should delete it.
We need to rethink the `v2`-API either way, as we want to be able to create apps without uploading an app at the same time as well. So in my mind the `v2`-API should warrant more changes than this - not just API-facing but overall code structure.

mediremi: >  There's some abstractions that are used for the Organisation that I would like to see be used more, and see how it performs in a more complex case like this

I've tried to base the create handler on [the create organisation handler](https://github.com/dhis2/app-hub/blob/0075932ebae37d852599c77f69e5d50385a817e6/server/src/routes/v2/organisations.js#L124-L143) so I'm not sure what other abstractions used for organisations I should be using but feel free to push the changes you're thinking of to this PR.

I've tried to limit the backend changes I've made in this PR due to i) the large diff, and ii) the focus being on the frontend - hence my original update to the create app handler being pretty much copy-paste from the v1 with only the changes that I required for the frontend to work.

> I do think that the main changes you've done could be done directly in the v1 API, as that needs fixing for the user-problem either way. I also don't think we should delete it.

I'll move the v2 create app handler (as well as the tests) back to v1 :+1:

Birkbjo:
I also noticed that the new design is pretty restrictive in its responsiveness. The old design is pretty mobile-friendly. Probably not a big issue since the main use-case would be desktop, but would be great if at least the public-facing design was more mobile friendly.

amcgee:
Screen Shot 2021-04-26 at 15 27 25

1. 
Some thoughts on the design here - it seems busy to me.  Do we really need so many "edit" buttons for each field?  Wouldn't a single "edit" button be enough since they all link to this "edit" page?

Screen Shot 2021-04-26 at 15 28 19

2. I don't know that we want to support and certainly not encourage "editing" a version.  Once a version is created it should be immutable, in my opinion - maybe it could be "deleted" (this should actually just hide the version, not delete it from the database), but shouldn't be editable.

CC @cooper-joe

client/src/components/Versions/Versions.js
```diff
@@ -0,0 +1,86 @@
...
+            />
+            {filteredVersions.length > 0 ? (
+                <VersionsTable
+                    versions={filteredVersions}
+                    renderEditVersionButton={renderEditVersionButton}
+                    renderDeleteVersionButton={renderDeleteVersionButton}
+                />
+            ) : (
+                <em className={styles.noVersions}>
+                    There are no compatible versions matching your criteria
```
cooper-joe, line 73: I suggest changing to `No versions found. Try adjusting your search.`

client/src/pages/UserOrganisations/UserOrganisations.js
```diff
@@ -0,0 +1,132 @@
...
+                                <Button primary large>
+                                    Create your first organisation
+                                </Button>
+                            </Link>
+                        </div>
+                    </>
+                )}
+                {organisations.length > 0 &&
+                    filteredOrganisations.length === 0 && (
+                        <em>No organisations match your search criteria.</em>
```
cooper-joe, line 114: I suggest `No organisations found. Try adjusting your search.`

client/src/pages/UserOrganisations/UserOrganisations.js
```diff
@@ -0,0 +1,132 @@
...
+                />
+            </div>
+            <Card className={styles.card}>
+                <h2 className={styles.cardHeader}>Your organisations</h2>
+                {organisations.length === 0 && (
+                    <>
+                        <em>You are not a member of any organisations.</em>
+                        <div style={{ marginTop: 8 }}>
+                            <Link to="/user/organisations/new" tabIndex="-1">
+                                <Button primary large>
```
cooper-joe, line 105: I suggest button `basic`, rather than `primary`.

cooper-joe (dismissed):
I've reviewed the design, adding suggestions to adjust labels and component usage. Where possible, I've commented directly. There are a few points that relate generally:

## Section: Your Apps

### Upload new version

- Would it be possible to have the title of this page reflect the app that is being updated? Using *Create new app version* could be confusing. I suggest *Upload new version: $appName* as the page title instead.

### Create new app

- Cannot use the *Select* menus with the keyboard. Opening the menu makes the page scroll down, and I am not able to navigate the menu or select an option. (Firefox) (This may be a UI issue? But logging here to check).
- Would it be possible to have a loading state once the user clicks the submit button? I had to wait a while (bad network), and it wasn’t clear what was happening. We could change the button text to “Submitting app”, or use a loading spinner somehow. Let me know our options here and we can discuss.

### My app details

- I agree with the comment @amcgee raised above, I don’t think each value needs it’s own edit button. Instead, I suggest adding a single button at the end of the section, “Edit app details”, for example.

## Section: Organisation

- All modal views should be using right-aligned buttons, as defined in the [modal usage guidelines](https://github.com/dhis2/design-system/blob/master/molecules/modal.md#1-task-based). Most modals in this section have the confirm buttons aligned left, which is not the correct usage.

- I get a *conflict* error when trying to create multiple organisations. The error is no more specific than *conflict,* so I’m not sure what the issue is. Is having two organisations with the same email restricted? I have been able to do it, but get constant errors when trying again.

mediremi:
> Cannot use the Select menus with the keyboard. Opening the menu makes the page scroll down, and I am not able to navigate the menu or select an option. (Firefox) (This may be a UI issue? But logging here to check).

Yep it's a UI component issue and I have updated https://github.com/dhis2/notes/discussions/238 to include `SingleSelect`.

> Would it be possible to have a loading state once the user clicks the submit button? I had to wait a while (bad network), and it wasn’t clear what was happening. We could change the button text to “Submitting app”, or use a loading spinner somehow. Let me know our options here and we can discuss.

This is definitely possible - I guess we could combine the updated text and spinner in the button?
image

> I get a conflict error when trying to create multiple organisations. The error is no more specific than conflict, so I’m not sure what the issue is. Is having two organisations with the same email restricted? I have been able to do it, but get constant errors when trying again.

I think this error is either due to a naming conflict (organisation names must be unique) or duplicate emails.

client/src/pages/UserAppUpload/UserAppUpload.js
```diff
@@ -0,0 +1,424 @@
...
+                                find your app.
+                            </p>
+                            <ReactFinalForm.Field
+                                required
+                                name="logo"
+                                buttonLabel="Upload a logo"
+                                accept="image/*"
+                                component={FileInputFieldFF}
+                                className={styles.field}
+                                validate={hasValue}
```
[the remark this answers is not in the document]
mediremi: The backend checks that the mime type starts with `image/` so pretty much any image file format is supported. Upload size is a bit trickier since we set a maximum multipart form size of 20MB, which for this page would include the bundle size. When uploading images separately the maximum size is 20MB (which I think we should reduce to ~5MB) so it's not exactly intuitive for users. I guess we could put something like "2MB maximum file size" in the help text as a kind of soft limit?

mediremi:
> I don't know that we want to support and certainly not encourage "editing" a version. Once a version is created it should be immutable, in my opinion - maybe it could be "deleted" (this should actually just hide the version, not delete it from the database), but shouldn't be editable.

I agree that versions should be immutable and that deleting versions should be replaced with hiding them. In this PR I'll just remove the edit version button and will defer version management changes due to the backend work that would be required (I'm trying to keep this PR as frontend focused as possible due to the large diff).

client/src/pages/UserAppUpload/UserAppUpload.js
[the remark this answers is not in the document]
cooper-joe: Yeah, I think that would work. I think a help text with the common image formats would help avoid the most common mistake (trying to upload a PDF). So in that case I'd go for `Only .jpg, .png and .gif files, 2MB max size`.

cooper-joe:
> This is definitely possible - I guess we could combine the updated text and spinner in the button?

That works for me. (I think I'll work on a `loading` state for buttons so we can standardize it in the component, this is fine for now).

> >    I get a conflict error when trying to create multiple organisations. The error is no more specific than conflict, so I’m not sure what the issue is. Is having two organisations with the same email restricted? I have been able to do it, but get constant errors when trying again.

> I think this error is either due to a naming conflict (organisation names must be unique) or duplicate emails.

Strange. I was able to create two organisations with the same email after a few tries.

client/src/pages/UserApp/ScreenshotsCard/ScreenshotsCard.js
```diff
@@ -0,0 +1,106 @@
...
+    return (
+        <Card className={sharedStyles.card}>
+            <h2 className={sharedStyles.cardHeader}>Screenshots</h2>
+            {screenshots.length > 0 ? (
+                <Screenshots
+                    screenshots={screenshots}
+                    onDelete={handleScreenshotDelete}
+                />
+            ) : (
+                <em>This app has no screenshots</em>
```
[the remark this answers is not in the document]
mediremi: Like this?
image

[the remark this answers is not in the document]
cooper-joe: Yeah, looks good 👍

Birkbjo:
> > I think this error is either due to a naming conflict (organisation names must be unique) or duplicate emails.
> 
> Strange. I was able to create two organisations with the same email after a few tries.

Emails are definitely not unique, but the organisation name is. We should use the backend-error message in that case @medi . Seems like the `useErrorAlert` uses the `error`-property instead of the `message`-prop? 

Example message from the server: `Organisation already exists or is too similar to existing organisation (DHIS2)"`

mediremi:
> Seems like the useErrorAlert uses the error-property instead of the message-prop?

The issue came from the way the two API clients were converting server error responses into JavaScript errors and is fixed by commit https://github.com/dhis2/app-hub/pull/458/commits/185ece7011a5dec47571fa53adc05ec9ec44e359.

Birkbjo (approved):

varl:
Yeehaw !! :rocket:

PR #463, opened 2021-04-22
fix: user email migration

Birkbjo:
Relates to https://jira.dhis2.org/browse/HUB-105 and is an implementation of the proposal here https://github.com/dhis2/notes/discussions/260


client/src/components/verify/OrganisationInvitationCallback.js
```diff
@@ -0,0 +1,51 @@
...
+import { Redirect } from 'react-router-dom'
+import { useQueryParams, StringParam } from 'use-query-params'
+import { acceptOrganisationInvitation } from '../../api/api'
+
+const OrganisationInvitationCallback = () => {
+    const [queryParams] = useQueryParams({
+        invitationToken: StringParam,
+    })
+
+    const { invitationToken } = queryParams
```
mediremi, line 11:
```suggestion
    const [invitationToken] = useQueryParam('invitationToken', StringParam)
```

server/src/routes/v1/apps/handlers/editImage.js
```diff
@@ -36,11 +36,16 @@ module.exports = {
...
         const currentUser = await getCurrentUserFromRequest(request, db)
-        const appDeveloperId = await getAppDeveloperId(appId, db)
-
+        const appsUserCanEdit = await getOrganisationAppsByUserId(
+            currentUser.id,
+            db
+        )
+        const userCanEditApp =
+            appsUserCanEdit.filter(app => app.app_id === request.params.appId)
+                .length > 0
```
mediremi, line 45:
```suggestion
        const userCanEditApp =
            appsUserCanEdit.map(app => app.app_id).includes(request.params.appId)
```

server/src/routes/v2/organisations.js
```diff
@@ -327,4 +330,132 @@ module.exports = [
...
+            validate: {
+                params: Joi.object({
+                    token: Joi.string().required(),
+                }),
+            },
+        },
+        handler: async (request, h) => {
+            const { id } = await getCurrentUserFromRequest(request)
+            const { db } = h.context
+            const token = request.params.token
```
mediremi, line 429:
```suggestion
            const { token } = request.params
```

server/src/server/env-config.js
```diff
@@ -17,6 +17,7 @@ const config = {
     server: {
         port: process.env.PORT || 3000,
         host: process.env.HOST || 'localhost',
+        jwtSecret: process.env.INTERNAL_JWT_SECRET,
```
mediremi, line 20:
```suggestion
        jwtSecret: getRequiredEnv(process.env.INTERNAL_JWT_SECRET),
```

server/src/server/env-config.js
```diff
@@ -25,6 +26,11 @@ const config = {
         level: process.env.NODE_ENV !== 'test' ? 'info' : 'error',
         prettyPrint: process.env.NODE_ENV !== 'test',
     },
+    aws: {
+        accessKeyId: process.env.AWS_ACCESS_KEY_ID,
+        secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY,
```
mediremi, line 31:
```suggestion
        accessKeyId: getRequiredEnv(process.env.AWS_ACCESS_KEY_ID),
        secretAccessKey: getRequiredEnv(process.env.AWS_SECRET_ACCESS_KEY),
```

Birkbjo: I might actually remove these from the config as these are especially sensitive,  they are not used anywhere as SES-API read them directly from env and we do not want to accidentally print them out.

server/src/server/env-config.js
Birkbjo: This is technically not required as you could use `sendmail` in production

server/src/services/EmailService.js
```diff
@@ -0,0 +1,75 @@
...
+const FROM_INFO = {
+    name: 'DHIS2 App Hub',
+    address: 'apphub@dhis2.org',
+}
+class EmailService extends Schmervice.Service {
+    constructor(server, schmerviceOptions, serviceOptions) {
+        super(server, schmerviceOptions)
+
+        if (!serviceOptions.transport) {
+            server.logger.warn(
```
mediremi, line 17: I don't think `sendmail` as a fallback transport is what we want - it might be better to use a `console.log`/`debug` only transport for development that just prints out the email contents while for production we should enforce `serviceOptions.transport`.

mediremi: Here's a minimal development transport:

```
let transport = {
    name: 'developmentTransport',
    send: (mail, callback) => {
        const input = mail.message.createReadStream()
        console.log('Sent email:')
        input.pipe(process.stdout)
        input.on('end', function () {
            callback(null, true)
        })
    }
};
```

server/src/services/organisation.js
```diff
@@ -177,6 +178,28 @@ const hasUser = async (id, userId, knex) => {
     return hasUser.length > 0
 }
 
+const generateInvitationToken = ({ organisation, user, secret }, emailTo) => {
+    secret = process.env.INTERNAL_JWT_SECRET
+    if (!secret) {
+        throw new Error('No secret setup, set INTERNAL_JWT_SECRET')
+    }
```
mediremi, line 185: Since we want errors ASAP instead of at runtime, switching to this + `getRequiredEnv(process.env.INTERNAL_JWT_SECRET)` should make our lives easier:
```suggestion
const generateInvitationToken = ({ organisation, user }, emailTo) => {
    const secret = process.env.INTERNAL_JWT_SECRET
```

mediremi:
What do you think about something like this for simple rate limiting:

```javascript
const LRU = require('lru-cache')

class RateLimiter {
    constructor({ maxPerHour }) {
        const oneHourMs = 1000 * 60 * 60

        this.cache = new LRU({
            maxAge: oneHourMs
        })
    }

    do(fn, key) {
        const count = this.cache.get(key) || 0
        if (count < maxPerHour) {
            this.cache.set(key, count + 1)
            fn()
        } else {
            throw new Error('Rate limit exceeded')
        }
    }
}


class EmailService {
    constructor() {
        // ...

        const maxEmailsPerHour = 10
        this._rateLimiter = new RateLimiter({ maxPerHour: maxEmailsPerHour })
    }

    async sendOrganisationInvitation() {
        // ...

        return this._sendMail(sendTemplate, fromUserId)
    }

    _sendMail(email, key) {
        this._rateLimiter.do(() => {
            this.transporter.sendMail(email)
        }, key)
    }
}

```

mediremi:
I've tested the invitation/accepting functionality and it works great - way better than the previous UX of having to have signed in at least once. Also good to see the migration from `app.developer.name` and `app.developer.email` to `app.contactEmail` :+1:

Birkbjo:
> What do you think about something like this for simple rate limiting:
```

That would be great, but I wonder if it would be worth to look into something like https://github.com/wraithgar/hapi-rate-limit

mediremi:
> I wonder if it would be worth to look into something like https://github.com/wraithgar/hapi-rate-limit

Looks good let's try to integrate it once this PR is merged :+1:

mediremi (approved):

server/src/services/EmailService.js
Birkbjo: Ive replacted it with a JSONTransport for testing.

PR #466, opened 2021-04-26
feat: verify metadata

mediremi:
Closes https://jira.dhis2.org/projects/HUB/issues/HUB-109


Birkbjo (approved):
LGTM. Should probably update with some more specific tests, but we can introduce that with the `v2`-API.

PR #467, opened 2021-04-28
feat: backend changes required by redesign

mediremi:
Backend changes cherry picked from https://github.com/dhis2/app-hub/pull/458.


server/src/data/updateApp.js
```diff
@@ -28,6 +28,15 @@ const paramsSchema = joi
     })
     .options({ allowUnknown: true })
 
+const isValidSourceUrl = sourceUrl => {
```
Birkbjo, line 31: This could probably be done through Joi with, see https://joi.dev/api/?v=17.4.0#stringurioptions . Using `['http', 'https'` as URI-schemes.

server/src/services/app.js
```diff
@@ -0,0 +1,104 @@
...
+    createAppVersion,
+    createLocalizedAppVersion,
+    addAppVersionToChannel,
+    addAppMedia,
+} = require('../data')
+
+exports.create = async (
+    { userId: currentUserId, organisationId, appType, status },
+    db
+) => {
```
Birkbjo, line 13: What's the reason for not using a transaction here? We need this to be a transaction so if something fails we have orphan tuples in the DB. Please us `db.transaction(trx =>)` to compose these. Either in the `create`-service or in the handler and pass that as the `db`-parameter to this function.

server/src/data/updateApp.js
```diff
@@ -24,19 +24,13 @@ const paramsSchema = joi
         languageCode: joi
             .string()
             .max(2)
+            .uri({
```
Birkbjo: Think you meant to do this on `sourceUrl` and not `languageCod` 😛

Birkbjo: Or you're testing me...

Birkbjo (approved):

PR #468, opened 2021-04-29
fix: encode strings with punycode before generating slug

mediremi:
We use [slugify](https://www.npmjs.com/package/slugify) to generate slugs for app versions and organisation names. Unfortunately, it strips most unicode characters, resulting in incorrect slugs.


server/src/utils/slugify.js
```diff
@@ -0,0 +1,4 @@
+const punycode = require('punycode')
```
Birkbjo, line 1: Seems like the core `punycode`-module is being deprecated: https://nodejs.org/api/punycode.html . Should probably add it as a dependency and import with `punycode/` https://github.com/bestiejs/punycode.js?PHPSESSID=5790fa7ffb5d909fb63c31c0f6571035

mediremi: Good catch!

Birkbjo (approved):

PR #471, opened 2021-05-03
fix: use punycode.toASCII instead of punycode.encode

mediremi:
In order to aid decoding, `punycode` appends a hyphen to the end of encoded strings. As we do not need to decode our slugs, using `toASCII` fits our use case better.


Birkbjo (approved):
LGTM!

PR #472, opened 2021-05-03
feat: add favicons

mediremi:


server/src/plugins/staticFrontendRoutes.js
```diff
@@ -1,22 +1,31 @@
 const path = require('path')
 
+const staticDir = path.join(__dirname, '..', '..', 'static')
+
+const faviconRoutes = [
```
Birkbjo, line 5: We could probably also just update the `href` in the `link`-in the headers above. Don't know if thats better or not...

mediremi: The reason why I had to make a new `public` folder and then add the routes instead of updating the `<link>` tags and reuse the static folder is because of the way our build script currently works: https://github.com/dhis2/app-hub/blob/master/Dockerfile#L13-L14. The `static` folder is wiped and overriden so we would have to somehow get webpack to copy these favicons during production build but that seemed more complicated then this solution.

mediremi: Do you think we should do something like map any route that matches something like `/favicon/:filename` to reading that file from the public dir instead? Seems cleaner than hardcoding like I've done.

Birkbjo: > Do you think we should do something like map any route that matches something like /favicon/:filename to reading that file from the public dir (and updating the hrefs) instead? Seems cleaner than hardcoding like I've done.

Something like that would be great, yeah. I do think that we should have a way to include static-files though. It doesn't seem like the `static`-folder is actually wiped? Just "overridden" by the client-build files. We could remove the `static`-entry from `gitignore` and have the favicons there, right? 

Another way, like you mentioned,  is to use something like https://webpack.js.org/plugins/copy-webpack-plugin/ and have the favicons as part of the client-build process - that way the favicons would work when running from webpack-dev-server as well.

Birkbjo: I updated webpack and tried to integrate the second option here: https://github.com/dhis2/app-hub/pull/475/files

Might want to split it, but we could just merge it as a regular merge as well.

mediremi: I'll close this PR and approve yours :+1:

mediremi:
Closing in favour of https://github.com/dhis2/app-hub/pull/475

PR #473, opened 2021-05-05
feat: add version API

mediremi:


Birkbjo (approved):
So simple yet effective!

PR #491, opened 2021-05-25
chore(release): release redesign

Birkbjo:
Releases the `UI-redesign`.

Note that this adds a DB-migration, and is thus not risk-free.

We should merge #492 before we release this.


mediremi (approved):
:rocket:

varl (approved):
Agree that the other PR needs to be merged first, feel free to go ahead after that.

PR #518, opened 2021-06-22
fix: hide channels filter

Birkbjo:


mediremi (approved):
I think having the feature be 'positive' and named `channels` instead of `hideChannels` might make more sense but I'm not going to block approval over that :+1:

varl (approved):
Agree with @mediremi on the name, and also on the "not gonna block" over that.

PR #521, opened 2021-07-05
fix(verifybundle): ignore name-check for dashboard_widget

Birkbjo:
See https://jira.dhis2.org/browse/HUB-123

The check might look a bit weird, but I did not want to introduce a requirement for `manifest.appType`, even though it should be in the manifest. It's either that or a check like `if (manifest.name !== appName) && !manifest.appType || (manifest.appType && manifest.appType !== 'DASHBOARD_WIDGET')`. If you have a better solution please give feedback.

### DIFF
--- a/server/src/security/verifyBundle.js
+++ b/server/src/security/verifyBundle.js
@@ -14,7 +14,7 @@ const checkManifest = ({ manifest, appId, appName, version, canBeCoreApp }) => {
         throw new Error('Manifest App Hub ID does not match app ID')
     }
     if (manifest.name !== appName) {
-        if (manifest.appType && manifest.appType === 'DASHBOARD_WIDGET') {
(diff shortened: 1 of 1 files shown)


mediremi (approved):
Having the check as a nested if is fine with me :+1: 

However isn't `manifest.appType && manifest.appType === 'DASHBOARD_WIDGET'` equivalent to `manifest.appType === 'DASHBOARD_WIDGET'`?

Birkbjo:
> Having the check as a nested if is fine with me 👍
> 
> However isn't `manifest.appType && manifest.appType === 'DASHBOARD_WIDGET'` equivalent to `manifest.appType === 'DASHBOARD_WIDGET'`?

Yes, of course. I wrote the condition as `!==` first, so that just remained. Thanks!

PR #522, opened 2021-07-06
feat: add coreApp property

Birkbjo:
Adds a property for identifying a core app.

The `core_app`-field is read from the manifest and set if it is true. In that sense this is the first information we are automatically getting from the manifest/d2.config. Note that the uploader must be a manager to be able to set this. 
I've also implemented a simple override for this when editing an application. And you have to be a manager for this property-change to have an effect. 

I'm a bit unsure about how we should show this to users, currently I have a simple tag for this:

image
Could potentially have the DHIS2 logo instead? Feedback welcome.

### DIFF
--- a/server/src/data/getApps.js
+++ b/server/src/data/getApps.js
@@ -43,10 +43,10 @@ const getApps = (
             }
 
             if (coreApp) {
-                builder.where('organisation', 'DHIS2')
+                builder.where('core_app', true)
             }
             if (coreApp === false) {
-                builder.whereNot('organisation', 'DHIS2')
+                builder.whereNot('core_app', false)
             }
 
             if (query) {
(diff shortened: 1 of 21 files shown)


Birkbjo:
I've also added a way to sort by 'Core apps' only, but I'm not sure if that is something we want to include.

server/src/data/getApps.js
```diff
@@ -43,10 +43,10 @@ const getApps = (
             }
 
             if (coreApp) {
```
mediremi, line 45: Now that we have a `core_app` boolean column I guess we can do:

```
if (typeof coreApp === 'boolean') {
    builder.where('core_app', coreApp)
}
```

mediremi (approved):
LGTM :heavy_check_mark: 

It's nice to have this as a proper column instead of the previous method of checking if the org was `DHIS2` :+1:

PR #528, opened 2021-08-18
feat: AppVersion API

Birkbjo:
Implements an AppVersions API that can be used to paginate and filter versions of an app.
This uses the new architecture, and takes full advantage of the queryFilter-plugin and pagination plugin, as well as the `executeQuery`-abstraction. 

Appversion is implemented as a Schmervice service. This is not strictly necessary as we do not have state or initialization. But I do like the approach, and it makes implementing caching pretty trivial if that is needed down the line.

### DIFF
--- a/server/src/server/init-server.js
+++ b/server/src/server/init-server.js
@@ -125,6 +125,7 @@ exports.init = async (knex, config) => {
     await server.register({
         plugin: pagination,
     })
+    server.method(appVersionMethods)
 
     await server.start()
 
--- a/server/src/server/init-server.js
+++ b/server/src/server/init-server.js
@@ -16,6 +16,7 @@ const errorMapper = require('../plugins/errorMapper')
 const queryFilter = require('../plugins/queryFilter')
 const pagination = require('../plugins/pagination')
 const { createEmailService } = require('../services/EmailService')
+const { createAppVersionService } = require('../services/appVersion')
 
 exports.init = async (knex, config) => {
     debug('Starting server...')
@@ -90,6 +91,7 @@ exports.init = async (knex, config) => {
     await server.register(Schmervice)
 
     await server.registerService(createEmailService)
+    await server.registerService(createAppVersionService)
 
     await server.register({
         plugin: staticFrontendRoutes,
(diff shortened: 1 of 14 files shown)


server/src/services/appVersion.js
```diff
@@ -0,0 +1,70 @@
...
+            pager,
+            model: AppVersionModel,
+        })
+
+        console.log('rez', res)
+        return res
+    }
+}
+
+const createAppVersionService = (server, schmerviceOptions) => {
```
Birkbjo, line 58: We could've just done `server.register(AppVersionService)` class directly. But if we want to do some initialisation or add caching down the line, this approach of a "service factory" is more flexible.

server/src/server/init-server.js
```diff
@@ -90,6 +91,7 @@ exports.init = async (knex, config) => {
     await server.register(Schmervice)
 
     await server.registerService(createEmailService)
```
Birkbjo, line 93: We might want to do the initialisation of these services in the `apiRoutes`-plugin. Then we wouldn't have to call `request.services(true)`, as it would be the same plugin that registered the schmervice-services. It can be argued that these services is not really a part of the api-routes though, so I've kept it in `init-server.js`.

server/src/services/appVersion.js
```diff
@@ -0,0 +1,69 @@
...
+        .select(
+            'app_version.id',
+            'app_version.version',
+            knex.ref('app_version.app_id').as('appId'),
+            knex.ref('app_version.created_at').as('createdAt'),
+            knex.ref('app_version.source_url'),
+            knex.ref('app_version.demo_url'),
+            knex.ref('channel.name').as('channel'),
+            knex.ref('ac.min_dhis2_version'),
+            knex.ref('ac.max_dhis2_version')
```
mediremi, line 27: If we add some `.as()` calls here then we could remove the `.rename()` calls in the model/definition file

```suggestion
            knex.ref('app_version.source_url').as('sourceUrl'),
            knex.ref('app_version.demo_url').as('demoUrl'),
            knex.ref('channel.name').as('channel'),
            knex.ref('ac.min_dhis2_version').as('minDhis2Version'),
            knex.ref('ac.max_dhis2_version').as('maxDhis2Version')
```

PR #543, opened 2021-09-06
fix(pagination): separate total count query

Birkbjo:
In the first iteration of pagination we used the window function `over()`, to do a somewhat performant and simple way to grab the total number of results in a query (before limit is applied). However this does not support `distinct`, which is needed in a `AppVersion` query. So whenever a `distinct` is used in a query, and whenever there are duplicate tuples, the total will be more than what we actually want.

In this PR I've switched out the window-function with a separate total-count query. In the initial implementation I argued against this due to performance reasons. But due to the constraints above, I've decided to implement it with a separate query, as it's much more flexible (which is how DHIS2 handles `total` as well). There's some performance issues with `total` in DHIS2, but that is with millions of entries and our database is not expected to grow to that level. The queries take single digit `ms` on my machine. 

Another approach would be to rewrite queries to not use distinct, but that complicates these queries.

### DIFF
--- a/server/src/query/executeQuery.js
+++ b/server/src/query/executeQuery.js
@@ -10,7 +10,8 @@ const debug = require('debug')('apphub:server:executeQuery')
  * @param {*} queryHelpers.model a "model" object that can be used to format the output.
  * @param {*} options Options object
  * @param {*} options.formatter a function with signature `function(result)` that overrides format-logic from model, should return formatted result
- */
+ * @param {Boolean} options.slice if true and pager is present, the result will be sliced in memory, without using a second SQL-query.
+ * */
 async function executeQuery(
     query,
     { filters, pager, model } = {},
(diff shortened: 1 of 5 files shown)


Birkbjo:
Test is failing due to a stupid mock. I will rewrite all the `executeQuery`-test to use actual knex-queries instead of mocks.

server/src/query/executeQuery.js
```diff
@@ -26,12 +27,20 @@ async function executeQuery(
     debug('Executing query: ' + query.toString())
     const rawResult = await query
     let result = rawResult
+    let totalCount = result.length
+
+    if (pager) {
+        const countQuery = pager.getTotalCountQuery(query)
```
mediremi, line 33: Am I understanding this right in that we are running the same query twice:
1. With a `limit` and `offset`, and whose results are what we return
2. Without `limit` and `offset` but with `count`

Could we instead run it once without `limit` and `offset` and apply those ourselves afterwards?

```js
const applyOffset = (result, offset) => {
  return result.slice(offset)
}

const applyLimit = (result, limit) => {
  return result.slice(0, limit)
}

const result = await query
return {
  count: result.length,
  result: applyOffsetAndLimit(result, { offset, limit })
}
```

Birkbjo: Yes, because we are doing the offset in the SQL-query. Of course we could do this in JS, and we do have that implemented in the decorated `pagination`-implementation. However that means we need all results in memory. It's a tradeoff, and I think in our case the result-set should stay small enough that it probably doesn't really matter. But I do think the memory usage would become a problem before a second query would.

I can implement a method for this in the Pager-class, and then we can decide between both for different use cases.

mediremi: No need for a separate method I was just wondering what the tradeoffs were :+1:

mediremi (approved):

PR #551, opened 2021-09-15
fix(appversion): fix version filtering

Birkbjo:
We allow null and empty strings for `maxDhisVersion`, and the previous filtering implementation would just do a naive `where`, so it would not return any results.

This implements a generic way to include "empty" results for a filter through an option `includeEmpty` to `applyOneToQuery`. Empty is defined as `null`, the empty string `''` or a string with just spaces.

Implements a simple way to compare version numbers in SQL. Note that this is not fully `semver` compatible, as it only supports simple  version-numbers (no prelease, beta, etc), however that should be fine for now as we only use it for DHIS core version. There is a [semver postgres-plugin](https://pgxn.org/dist/semver/), but that is more work to implement. 

Also includes `gte` (greater or equal) to the list of allowed operators for filtering.

### DIFF
--- a/server/src/services/appVersion.js
+++ b/server/src/services/appVersion.js
@@ -20,6 +20,7 @@ const getAppVersionQuery = knex =>
             'app_version.version',
             knex.ref('app_version.app_id').as('appId'),
             knex.ref('app_version.created_at').as('createdAt'),
+            knex.ref('app_version.updated_at').as('updatedAt'),
             knex.ref('app_version.source_url').as('sourceUrl'),
             knex.ref('app_version.demo_url').as('demoUrl'),
             knex.ref('channel.name').as('channel'),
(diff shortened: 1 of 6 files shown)


server/src/services/appVersion.js
```diff
@@ -46,6 +48,17 @@ class AppVersionService extends Schmervice.Service {
             })
         }
 
+        // null-values are allowed for maxDhisVersion, so include these if filter is present
+        if (filters.getFilter('maxDhisVersion')) {
```
mediremi, line 52: We could push the `if` check to `applyVersionFilter`, allowing us to just do

```js
filters.applyVersionFilter(query, 'maxDhisVersion', {
    includeEmpty: true,
})
filters.applyVersionFilter(query, 'minDhisVersion')
```

Birkbjo: Interesting thought. I'm not sure that coupling the filter-names/logic to that function is worth it, as we could use it elsewhere.

Birkbjo: Nvm, I'm stupid. I understand what you mean - will add that instead!

mediremi (approved):
LGTM :heavy_check_mark:

PR #560, opened 2021-10-01
fix: add super agent proxy

Birkbjo:
`rest-facade` ( a dependency of `auth0`) moved `superagent` from dependency to peerDependency without a major version bump. This causes the resolved version to not include this dependency and the server fails on startup due to `no module found`. 

This should hopefully be fixed upstream, and others have also been affected by this: https://github.com/ngonzalvez/rest-facade/issues/60 and https://github.com/auth0/node-auth0/issues/663 .

If this is not fixed in either `rest-facade` or `auth0`  by monday, we should merge this. Otherwise we can just do a version bump when that happens.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -47,6 +47,7 @@
         "rimraf": "^3.0.1",
         "semver": "^7.3.2",
         "slugify": "^1.6.0",
+        "superagent": "^6.1.0",
         "uuid": "^8.3.2"
     },
     "devDependencies": {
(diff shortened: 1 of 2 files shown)


ngonzalvez:
Hi, this is fixed in `v1.16.1` of `rest-facade`. Please confirm whether the issue is resolved.

Birkbjo:
Closing due to being fixed in both our pipeline and a new version of `rest-facade`, thanks for the quick fix @ngonzalvez !
Small explanation as to how this propagated to the staging server below:

This issue was a combination of breaking change by `rest-facade` and our publish pipeline ignoring the `yarn.lock`-file. The latter is the reason a broken dependency managed to be published and crash the staging-server upon start (it rolled back of course). 
This happened due to  a bug in the `Dockerfile`. The `yarn.lock`-file was not copied correctly to the image before running `yarn install`, only the server-folder was copied, and the `yarn.lock`-file is in the root... And without the lock-file, most (unless locked) dependencies have always resolved to the latest minor versions. 
This was a good discovery, and probably would have bitten us down the line anyway, so it's great that we noticed this.

But as @varl said best:
> that it's worked for so long is a true testament to the stability of the dependencies we have chosen.
_looks at the bright side of life_

PR #593, opened 2021-11-27
fix(filter): properly include version if maxDhisVersion is empty

Birkbjo:
Empty `maxDhisVersion` (`''`) should now properly be included when filtering on `maxDhisVersion`.
This was probably not caught due to me reading somewhere that `nullif('', ' ')` should return true (with any blank string), but that is not the case in postgres- which makes sense. And most tests used `lte`, which included the empty-string regardless of the `or nullif`-case. 

Also added some more tests and related seeds.

On a slightly related note, we should've some better constraints/checks for `maxDhisVersion` and `minDhisVersion`, as currently they allow anything to be stored. At least we should protect against this in the application-layer/Joi-validation.

### DIFF
--- a/server/src/utils/Filter.js
+++ b/server/src/utils/Filter.js
@@ -125,7 +125,7 @@ class Filters {
             query.where(builder => {
                 builder.where(nameToUse, toSQLOperator(operator, value), value)
                 if (settings.includeEmpty) {
-                    builder.orWhereRaw(`nullif( ??, ' ') is null`, nameToUse)
+                    builder.orWhereRaw(`nullif( ??, '') is null`, nameToUse)
                 }
             })
             this.markApplied(fieldName)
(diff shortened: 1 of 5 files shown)


server/src/utils/Filter.js
```diff
@@ -125,7 +125,7 @@ class Filters {
             query.where(builder => {
                 builder.where(nameToUse, toSQLOperator(operator, value), value)
                 if (settings.includeEmpty) {
-                    builder.orWhereRaw(`nullif( ??, ' ') is null`, nameToUse)
+                    builder.orWhereRaw(`nullif( ??, '') is null`, nameToUse)
```
amcgee, line 128: Do we want to catch all empty strings?  If so we could use [`TRIM`](https://www.w3schools.com/sqL/func_sqlserver_trim.asp) (only trims spaces though)

```suggestion
                    builder.orWhereRaw(`nullif( trim( ?? ), '') is null`, nameToUse)
```

amcgee (changes requested):
Looks good!  One suggestion / question / concern

server/src/utils/Filter.js
Birkbjo: I originally included that, but figured we rather want to be more strict on the allowed format and do this in the application-layer with Joi-validation.

amcgee: OK sure!

amcgee (approved):
👍

PR #603, opened 2021-12-06
fix: hide apps in header-menu when not logged in

Birkbjo:
Hide apps-menu in header bar when not logged in.

Also renames "Your apps" to "My apps", and other related menu-items, as suggested by @germanviscuso .

### DIFF
--- a/client/src/components/Header/Header.js
+++ b/client/src/components/Header/Header.js
@@ -16,7 +16,7 @@ const NotLoggedInIcon = () => {
 
     return (
         <button className={styles.signInButton} onClick={loginWithRedirect}>
-            Sign in
+            Developer Sign in
         </button>
     )
 }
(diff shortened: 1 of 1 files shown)


client/src/components/Header/Header.js
```diff
@@ -65,12 +65,12 @@ const ProfileButton = () => {
...
             <Divider dense />
-            <DropdownMenuItemLink to="/user">Your apps</DropdownMenuItemLink>
+            <DropdownMenuItemLink to="/user">My apps</DropdownMenuItemLink>
             <DropdownMenuItemLink to="/user/organisations">
-                Your organisations
+                My organisations
             </DropdownMenuItemLink>
             <DropdownMenuItemLink to="/user/apikey">
-                Your API keys
+                My API keys
```
amcgee, line 73: This is probably fine for now to be consistent with `My Apps`, but interestingly GitHub uses `Your` in their profile menu...

Screen Shot 2021-12-06 at 14 28 00

amcgee: (API Keys probably should be in a general `Settings` page, eventually, but fine for now)

amcgee (approved):
Minor point about `My` vs `Your` but the main feature of hiding `Your Apps` when not logged in looks good to me.

@cooper-joe @varl @germanviscuso maybe we defer the renaming to `My` to a later change?  I'd also like to explore* using verbs in the main menu, i.e. `Explore` and `Manage`

client/src/components/Header/Header.js
Birkbjo: Sure, we can defer the renaming. It's not really important, and will probably make our other docs/guides inconsistent.

PR #621, opened 2022-01-21
fix: improve UX of invalid image filetype errors

mediremi:
- Sets the accepted filetypes for various `<input type="file">` inputs used across the app for image uploads to `image/jpeg,image/png`.
- Improves the error message shown to users when an attempt at uploading an invalid image file is made

## Before
image

## After
image

### DIFF
--- a/server/src/utils/validateMime.js
+++ b/server/src/utils/validateMime.js
@@ -5,7 +5,7 @@ const Path = require('path')
 const allowedImageMimeTypes = ['image/jpeg', 'image/png']
 const imageMetadataSchema = Joi.object({
     headers: Joi.object({
-        'content-type': Joi.string().valid(...allowedImageMimeTypes),
+        'content-type': Joi.string(),
     }).unknown(),
     filename: Joi.string(),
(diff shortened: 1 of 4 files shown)


server/src/utils/validateMime.js
```diff
@@ -5,7 +5,7 @@ const Path = require('path')
 const allowedImageMimeTypes = ['image/jpeg', 'image/png']
 const imageMetadataSchema = Joi.object({
     headers: Joi.object({
-        'content-type': Joi.string().valid(...allowedImageMimeTypes),
```
mediremi, line 8: `.valid(...allowedImageMimeTypes)` had to be removed in otherwise the error returned by hapi would be the one from `Joi.assert` instead of `Boom.badRequest`.

Birkbjo: Was going to comment that we could add a message with `.message()`, but since we are validating it below either way, it is kind of superfluous anyway.

Birkbjo (approved):

PR #646, opened 2022-04-08
feat: add download-counter

Birkbjo:


server/src/routes/v1/apps/handlers/getAppFile.js
```diff
@@ -86,6 +73,7 @@ module.exports = {
             `${item.app_id}/${item.version_id}`,
             'app.zip'
         )
+        appVersionService.incrementDownloadCount(item.version_id, knex)
```
Birkbjo, line 76: Should we await this? It's not actually necessary right?

mediremi: I don't think it's necessary and we wouldn't want a (potentially slow if the DB is under heavy read/write) DB write to block a file download

mediremi (approved):

PR #658, opened 2022-05-04
fix(version): revert v1-api version limit

Birkbjo:


amcgee (changes requested):

amcgee (approved):
Looks good, thanks for the quick work @Birkbjo !

PR #661, opened 2022-05-05
feat(app-version): add route for getting appVersion by id

Birkbjo:
https://dhis2.atlassian.net/browse/HUB-143


server/src/security/index.js
```diff
@@ -100,6 +100,7 @@ const getUserDecoration = function () {
         return {
             id: userId,
             name,
+            isManager: currentUserIsManager(this),
```
Birkbjo, line 103: Convenience prop instead of needing to import `currentUserIsManager` in routes.

server/src/query/Pager.js
```diff
@@ -20,12 +20,11 @@ const defaultPagingQuerySchema = Joi.object({
     pageSize: Joi.number().default(25).min(1),
 })
 
-const withPagingQuerySchema = joiSchema =>
-    Joi.alternatives().try(
-        defaultPagingQuerySchema.concat(joiSchema),
-        joiSchema
-    )
-const withPagingResultSchema = joiSchema =>
+const withPagingQuerySchema = (joiSchema) => {
```
Birkbjo, line 23: These changes are mostly for swagger docs to be generated correctly.

server/src/services/appVersion.js
```diff
@@ -33,16 +34,46 @@ const getAppVersionQuery = knex =>
...
+        }
+
+        return result
+    }
+
+    async findOne(id, knex) {
+        return this.findOneByColumn(id, { columnName: 'id' }, knex)
+    }
+
+    async findOneByVersion(version, knex) {
```
tomzemp, line 73: is this being used anywhere?

server/src/utils/Filter.js
```diff
@@ -1,6 +1,11 @@
 const debug = require('debug')('apphub:server:utils:Filter')
 const Joi = require('./CustomJoi')
-const { parseFilterString, toSQLOperator } = require('./filterUtils')
+const {
+    parseFilterString,
+    toSQLOperator,
+    isVersionOperator,
+    allOperatorsMap,
```
tomzemp, line 7: why not import `versionOperatorMap` if you're just using this within the body of the applyVersionFilter method?

server/test/routes/v2/appVersions.js
```diff
@@ -195,6 +195,52 @@ describe('v2/appVersions', () => {
...
+                    expect(v.version).to.be.a.string()
+                    const [, major] = v.minDhisVersion.split('.').map(Number)
+                    expect([28, 32]).to.include(major)
+                })
+            })
+
+            it('should handle multiple values for filter operators', async () => {
+                const request = {
+                    method: 'GET',
+                    url: `/api/v2/apps/${dhis2App.id}/versions?minDhisVersion=ne:2.28,2.32`,
```
tomzemp, line 224: this test made me wonder about `eq` operator, which when I looked at it seems to be functioning the same as `in` operator whereas I assume two values with `eq` operator would behave as `and` type join (and hence return nothing)?

tomzemp (approved):
Looks good to me 🎉 

I'm not familiar with the code base, so don't give myself highest marks in my review skills here 🙈, but it seemed like the changes were logical and were working as intended.

server/test/routes/v2/appVersions.js
Birkbjo: Yeah, that's right. Might not make sense in this case but the operators are supposed to be very generic, and could make sense in other situations.

server/src/routes/v2/appVersions.js
```diff
@@ -55,15 +93,36 @@ module.exports = [
...
-            return h.response(versions)
+            versions.map(setDownloadUrl)
+
+            return h.response(result)
         },
     },
 ]
+
+async function checkVersionAccess(version, request, db) {
+    const user = request.getUser()
```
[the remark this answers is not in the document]
Birkbjo: True. It's not expensive at all, just creates an object from already known values on the `request`-object. But it is unnecessary so I will move it, thanks!

amcgee (approved):
It looks like @tomzemp has some good points in his review - I had a quick look and it seems good, @Birkbjo @tomzemp if you'd like me to dive a bit deeper and test I could do it tomorrow

PR #663, opened 2022-05-24
feat: add markdown support for app description

Birkbjo:
Adds Markdown support for App description using [react-markdown](https://github.com/remarkjs/react-markdown).

This is currently only a client-side addition, as we are not doing any validation server-side. However it should be safe, since `react-markdown` is safe by default. 

I've added a simple `Preview`-tab when editing the description. I don't think a WYSWIG-editor is needed for now, since basic markdown is pretty simple. But it could be something we could improve upon later.
We should however probably mention that the description supports markdown, and maybe link to a resource?

I've set some styles to prevent abuse of eg. headers, so they cannot dominate other headers on the page.


client/src/pages/Apps/AppCards/AppCardItem/AppCardItem.js
```diff
@@ -32,12 +25,27 @@ const AppCardItem = ({ id, name, developer, type, description, images }) => {
             </header>
 
             <p className={styles.appCardDescription}>
-                {summarise(description)}
+                <Summarise>
+                    <ReactMarkdown
+                        allowedElements={['p']}
```
Birkbjo, line 30: This is used to remove tags like headers from the Summarised description on the front-page.

HendrikThePendric (commented):
@edoardo and @Birkbjo could you perhaps sync up and see if it would make sense to use our own `RichTextEditor` here? 

This component is not yet available in @dhis2/ui, but ready to get ported over from the line-listing-app. It's tricky for me to assess if it would make sense here, but I think you guys will able to do so.

amcgee:
Having a sentence indicating Markdown support would be a good, easy improvement here.

HendrikThePendric (approved):
Thanks for the feedback reg. the `RickTextEditor`. Seems there isn't that much overlap there, so just letting users type in a `textarea` makes sense.

I think @amcgee made a good point that we need to let the users know they can actually use markdown. We should probably add that to the `helpText` of the textarea field. I'm approving the PR, but would like to see a note added about this.

Also, very minor, but I find the chosen component names a little bit out of the ordinary. I think the main problem is that they read like verbs instead of nouns. I tend to use these rules when naming a component:
- A component is a thing, so use a noun
- Name the component after what it does
- If the component has a limited scope, prefix with that scope (in this case everything is under the app-description scope)

Following these rules I'd end up using these names:
`AppDescriptionSummary`
`AppDescriptionEditor`
`AppDescriptionTextarea`

Birkbjo:
> Also, very minor, but I find the chosen component names a little bit out of the ordinary. I think the main problem is that they read like verbs instead of nouns. I tend to use these rules when naming a component: 

I guess, and normally I would probably agree. But I just followed the already established convention in the app. As you can see it's divided into pages like `UserApp` and `UserAppEdit` - to differentiate the different pages - edit mode and non-edit mode. But I can change these. 

Regarding `Summarise` I agree, and this was just due to me rewriting a function to a component, and I should probably have called in `Summary`. But it's a generic component, so `AppDescriptionSummary` is too concise. 

Not sure what `AppDescriptionTextArea` would refer to? `WriteContent` and `PreviewContent` refer to the tab-names and I think the names for those are fine.

PR #697, opened 2022-11-15
fix(apps): fix sourceUrl being cleared when version is uploaded

Birkbjo:
Fixes `sourceUrl` being cleared in the UI when a new version is uploaded. This is done by making sure any app-version sourceUrl is used, not just the first being "collected". This is the easiest fix that I could come up with that does not need to run any migration to populate missing sourceUrls.

The reason for this is that `source_url` is in the `app_version`-table. It really should be in the "app"-table, because we should not have different `source_url`s for each version. I guess this was done in good spirit, making it possible to "migrate" source to another repository host between versions - but it's such an obscure use-case and results in tons of duplicated data.

I've not done any database fixes, because that is much more work and quite risky. I think we can do that once we introduce a "v2/apps"-API.

### DIFF
--- a/server/src/routes/v1/apps/formatting/convertAppsToApiV1Format.js
+++ b/server/src/routes/v1/apps/formatting/convertAppsToApiV1Format.js
@@ -4,7 +4,7 @@ const debug = require('debug')(
 const { MediaType } = require('../../../../enums')
 const getServerUrl = require('../../../../utils/getServerUrl')
 
-const convertDbAppViewRowToAppApiV1Object = app => ({
+const convertDbAppViewRowToAppApiV1Object = (app) => ({
     appType: app.type,
 
     status: app.status,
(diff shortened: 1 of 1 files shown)


server/src/routes/v1/apps/formatting/convertAppsToApiV1Format.js
```diff
@@ -100,9 +100,17 @@ const convertAll = (apps, request) => {
             })
         }
 
+        // some app-version may not have a version set
+        // sourceUrl really should be in 'app'-table, not 'app_version'
+        if (app.source_url && !currentApp.sourceUrl) {
+            currentApp.sourceUrl = app.source_url
+        }
```
Birkbjo, line 107: This is the actual fix, rest is prettier-styling.

Topener (approved):
LGTM! 

I haven't actually run the code locally, but I see nothing that _could_ break here.

PR #718, opened 2023-01-18
ci: remove extraneous .ebextensions dir and update EB env names

radnov:
The `.ebextensions` dir containing the `proxy.config` nginx configuration extension file is no longer needed after [migrating App Hub to Elastic Beanstalk on Amazon Linux 2](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/using-features.migration-al.html). The `HTTP` to `HTTPS` redirect is now handled by an Application Load Balancer (as outlined [here](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/configuring-https-httpredirect.html#configuring-https-httpredirect.redirect)).

This PR also updates the publish workflow Elastic Beanstalk environment names, as the current ones cannot be modified and we need to have the old and new environments running at the same time to eliminate downtime. Note that the EB environment names are only "internal" and will not affect the end user.


Birkbjo (approved):
Thanks a lot for working on this!

PR #721, opened 2023-02-02
fix: increase nginx client_max_body_size

radnov:
Fixing a regression made in https://github.com/dhis2/app-hub/pull/718

Note that the fix is already applied to apps.dhis2.org


Birkbjo (approved):
Thanks for the quick fix!

PR #737, opened 2023-05-16
feat(notifications): add slack webhook for new apps

Birkbjo:
Implements a `NotificationService` to be able to send notifications. This is designed to take multiple `messagers`, so we can eg. add a `EmailMessager` later. 
`SlackNotificationMessager` implements an `incoming webhook` to post a slack channel. I will configure the webhook-url to post to the `app-hub` channel.


Topener:
Based on what I understand from the different components this looks good, though I doubt I'd be the best person to review this. Maybe someone else from the frontend team is better suited to review

PR #743, opened 2024-12-02
chore: run prettier format on all codebase

kabaros:


kabaros:
@Birkbjo @amcgee - could we merge this? just running `yarn format` as it seems like there are a lot of files not formatted, and they cause the diff on my branch to be too big

amcgee (approved):
LGTM

PR #744, opened 2024-12-03
feat: expose plugins and changelog in the API and app

kabaros:
This adds functionality to show the changelog in the UI and visually differentiate plugins in the list of apps. Implements [HUB-156](https://dhis2.atlassian.net/browse/HUB-156) and [HUB-144](https://dhis2.atlassian.net/browse/HUB-144).

From a UX perspective, both these functionalities will be improved in upcoming tickets, but this PR sets the ground for how we do it, while exposing a very basic UI to test the functionality.

**After a first discussion, we decided to move `d2config` and `changelog` columns to the `app` table instead of `appeversion` - in addition, we added a `change_summary` column to versions table that would allow users to highlight what changed in a specific version. The `d2config` is taken from the uploaded zip file - while we attempt to get the change log from the zip file or the GitHub source url. Once we have the full change log - if we have it - we can display it in the UI in numerous ways, for now, we just implemented the simplest solution to show the full change log**
 
The workflow is as follows:

- When uploading a new version or a new app, we save the `d2.config.js`  (as a JSON type) and `CHANGELOG.md` to the `app` table.
   - If we're uploading a new version, we _patch_ the app entry with the `d2config` and `changelog` from the last version.
   - This means we only have the last version of these files - this is ok because this is the information we'd be interested in in this context, i.e. an app will be shown as a "plugin" if the latest d2config has a plugin entry, it doesn't matter if it did or did not have such an entry in the past. The same for the changelog - with the addition that we can infer all history from one (good) changelog and we also didn't want to have a massive text file saved with each version unnecessarily  
- An indicator `has_plugin` is added to the `app_view` view.  `has_plugin` inspects the JSON type and returns a flag whether a plugin entry point is found or not. This is returned to `/apps` endpoint and `/apps/:id`.
   - the whole `d2config` - rather than just the plugin info - is saved as it could be used to infer other information (i.e offline support, RTL support etc..).
   - if / when the entryPoints logic change (for example, to have multiple plugins), then the logic in the view can be updated.
- Both `d2.config` and `changelog.md` are not returned fully as part of apps or appversions endpoints - only an indicator that they exist, or inferred info (i.e. `has_plugins`)
 - A separate endpoint to get the changelog is added `/apps/:id/changelog`. It is separate to avoid returning a potentially large text file in the apps or apps/:id response. We only return an indicator, which the UI can use to show a "view changes" button, and when that is clicked, it calls the API and returns the changelog.
    - The changelog returned is for the app - that is all the app versions for that app. The UI, then, can make decisions based on that - for example, one changelog from a different version likely includes the changes from previous versions too. This minimises the need to have a script to backfill old versions change logs, as once a new version is added - it will contain  all the changes for previous versions too (at least for apps that follow conventional change logs, like our internal apps), but it needs to be limited as well potentially as more versions with changelogs are added
- We look for the changelog in the uploaded version zip file - in most cases (in our apps), it's not part of the bundle though. So we also  _try_ and get the changelog from the app github url if that exists.
   - This is pretty crude at the moment, for example, it works with changelogs in master branch only (might add a second check for main branch)
   - There is also a new field `change_summary` added to the version table where users can just highlight what changed in a specific version manually - this is not exposed yet in the UI.

## ToDO :
- I have a script to backfill previously added apps (by reading d2config and changelog) from GitHub - still undecided if we need to do that or not. With the current approach, we will get the changelog once a new version is uploaded, so it doesn't seem like this 
- [future] add ability to manually input a "what changed section" to the version info. This is the `change_summary_ field in the versions table, and can be used along the change log or instead of it. 

## UX
The UX, especially for the Changelog, is a placeholder - once we have the changelog info, we can parse and display as we want. I want a simple version in first, and that we agree on the database and API structure.

But here are some UX questions that came to mind while working on this:
- now that there is a "plugin" tag, it makes the "Standard" tag less obvious - shall we change that to a Tag too? what does "standard" even mean?
- how to filter plugins? right now, I added a "backdoor" where the user can search the word "plugin" and that will show all plugins
- how to word the fact that an app is a a plugin - technically it _has_ a plugin (or multiple plugins), but it is can also be an app
- ....

## Screenshots

image

image

image

[HUB-156]: https://dhis2.atlassian.net/browse/HUB-156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HUB-144]: https://dhis2.atlassian.net/browse/HUB-144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/src/components/Versions/VersionsTable/VersionsTable.js
+++ b/client/src/components/Versions/VersionsTable/VersionsTable.js
@@ -57,7 +57,6 @@ const VersionsTable = ({
             {changelogVisible && (
                 <ChangeLogViewer
                     appId={appId}
-                    modalShown={changelogVisible}
                     onCloseChangelog={onCloseChangelog}
                 />
             )}
--- a/client/src/components/Versions/VersionsTable/VersionsTable.js
+++ b/client/src/components/Versions/VersionsTable/VersionsTable.js
@@ -52,16 +52,15 @@ const VersionsTable = ({
         setChangelogVisible(false)
     }
 
-    // ToDO: we can infer change log from one change log
-    // const anyVersionHasChanges = !!versions.find((v) => v.hasChangelog)
-
     return (
         <>
-            <ChangeLogViewer
-                appId={appId}
-                modalShown={changelogVisible}
-                onCloseChangelog={onCloseChangelog}
-            />
+            {changelogVisible && (
+                <ChangeLogViewer
+                    appId={appId}
+                    modalShown={changelogVisible}
+                    onCloseChangelog={onCloseChangelog}
+                />
+            )}
             <Table className={styles.table}>
                 <TableHead>
                     <TableRowHead>
--- a/client/src/components/Versions/VersionsTable/VersionsTable.js
+++ b/client/src/components/Versions/VersionsTable/VersionsTable.js
@@ -14,6 +14,7 @@ import { useCallback, useEffect, useState } from 'react'
 import styles from './VersionsTable.module.css'
 import config from 'config'
 import { renderDhisVersionsCompatibility } from 'src/lib/render-dhis-versions-compatibility'
+import ChangeLogViewer from '../ChangeLogViewer'
 
 const { appChannelToDisplayName } = config.ui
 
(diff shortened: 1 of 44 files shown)


kabaros:
@cooper-joe this is a very early version of the changelog (+ highlighting plugins) in the App Hub. My plan is to follow up with UX improvements once the basics are working, but wanted to keep you in the loop (and also, maybe - even in this basic version - there are some UX wins that can be gained)

server/src/data/getApps.js
```diff
@@ -50,6 +50,9 @@ const getApps = (
                 builder.where((builder) => {
                     builder.where('name', 'ilike', `%${query}%`)
                     builder.orWhere('organisation', 'ilike', `%${query}%`)
+                    if (query.match(/plugin/i)) {
```
kabaros, line 53: an obscure way to find plugins - by searching for the exact word "plugin". A backdoor until we decide how/if we expose finding plugins in the UI.

server/src/data/patchApp.js
```diff
@@ -0,0 +1,43 @@
+const joi = require('joi')
+
+const paramsSchema = joi
```
kabaros, line 3: the existing updateApp did too many things and I didn't want to use it - although this in theory can be used to update any properties in the app object, for now, I am restricting, through the joi validation, it to only the d2config and changelog.

kabaros:
@amcgee @Birkbjo this is ready for review now

@Birkbjo is it possible to deploy this PR to staging?

kabaros:
created a `next` branch and changed the base branch for this PR - PR will get deployed to staging once merged to `next`

client/src/components/Versions/VersionsTable/VersionsTable.js
```diff
@@ -39,65 +40,97 @@ const VersionsTable = ({
...
-                                ).toLocaleDateString()}
-                            </span>
-                        </TableCell>
+        <>
+            {changelogVisible && (
+                <ChangeLogViewer
+                    appId={appId}
+                    modalShown={changelogVisible}
+                    onCloseChangelog={onCloseChangelog}
+                />
```
amcgee, line 62: ⚠️ (minor) - this will *both* hide/show using the `modalShown` prop AND mount/unmount with the conditional render.  We should do one or the other, but I don't think we need to do both.

server/src/routes/v1/apps/formatting/convertAppsToApiV1Format.js
```diff
@@ -15,6 +15,9 @@ const convertDbAppViewRowToAppApiV1Object = (app) => ({
 
     name: app.name,
     description: app.description || '',
+
+    hasChangelog: app.has_changelog,
+    hasPlugin: app.has_plugin,
```
amcgee, line 20: Should `hasPlugin` also expose the `pluginType` from `d2.config.js`?  That's an immediate win since we have both dashboard and capture plugin types available to day.

server/migrations/20241118143001_d2config_changelog_columns.js
```diff
@@ -0,0 +1,109 @@
...
+    `)
+}
+
+exports.down = async (knex) => {
+    await knex.schema.table('app_version', (table) => {
+        table.dropColumn('version_change_summary')
+    })
+
+    await knex.schema.table('app', (table) => {
+        table.dropColumn('d2config')
```
amcgee, line 63: Did we decide to move `d2config` to `appVersion` since it will change over time and it will be useful to keep a record of the d2config for each version...?

amcgee (commented):
Added a few comments.  In particular curious about these two points which I may be misremembering where we landed in earlier discussions:

1. Exposing not just hasPlugin but also pluginType in view and API
2. Storing d2config per version (so that we can track changes to app d2config over time)

server/migrations/20241118143001_d2config_changelog_columns.js
kabaros: done

server/src/routes/v1/apps/formatting/convertAppsToApiV1Format.js
kabaros: exposed plugin_type, and now the Plugin tag shows the plugin type if it exists

image

PR #746, opened 2024-12-16
feat: add organisation overview page

kabaros:
This PR adds an organisation overview page, to support the page, it:
- extends the call to `/organisation/:id` to optionally include the apps
- adds a description field to the organisation
   - only the owner of the organisation (the user who created it by default) can edit this description
 - Links to the new organisation page from App pages
 
A followup [ticket](https://dhis2.atlassian.net/browse/HUB-161) will expose ability to filter by organisation to the app.

implements [HUB-122](https://dhis2.atlassian.net/browse/HUB-122).

image

[apphub-organisation-page.webm](https://github.com/user-attachments/assets/09454684-d683-4468-91de-79b374bc669d)

[HUB-122]: https://dhis2.atlassian.net/browse/HUB-122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/src/pages/AppView/AppView.js
+++ b/client/src/pages/AppView/AppView.js
@@ -26,6 +26,7 @@ const HeaderSection = ({
     logoSrc,
     hasPlugin,
     pluginType,
+    organisationSlug,
 }) => (
     <section
         className={classnames(styles.appCardSection, styles.appCardHeader)}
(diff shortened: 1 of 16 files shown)


client/src/pages/AppView/AppView.js
```diff
@@ -24,14 +24,23 @@ const HeaderSection = ({
...
     >
         <AppIcon src={logoSrc} />
         <div>
             <h2 className={styles.appCardName}>{appName}</h2>
-            <span className={styles.appCardDeveloper}>by {appDeveloper}</span>
+            <span className={styles.appCardDeveloper}>
+                by{' '}
+                <a
+                    className={styles.link}
+                    href={`/organisation/${organisationSlug}/view`}
```
amcgee, line 39: Is it possible that `organisationSlug` will be null/undefined here leading to a broken link?  Or is it guaranteed to always be populated by the backend?

amcgee (approved):
Minor comment but LGTM!

client/src/pages/AppView/AppView.js
kabaros: it is guaranteed to not be null/undefined, but  the guarantee is happening at the application layer not the DB, there is no way that it could end up being null/undefined (we use slugify based on the name which is required) unless someone manually alters the DB ..

PR #747, opened 2024-12-19
feat: view changes between app versions based on the Changelog

kabaros:
implements [HUB-160 ](https://dhis2.atlassian.net/browse/HUB-160)

The PR adds the ability to view  an app change log:

- This is based on the CHANGELOG.md that is automatically generated for apps that use semantic-release (all of our apps)
- We will also be exposing a manual entry for each version `change_summary`, where the users can write whatever they want to describe what changed in a specific version
  - When that `change_summary` exists, it will be displayed first next to the entry (then any other changes from the changelog)

It also:

- changes the design of the app details page according to the new specs
- adds tests for the UI, and migrates from mocha to jest
- fixes some linting issues with css modules

**To test this:** Upload a new version of one of our apps (since these have a changelog generated) - this will update the app entry with the changelog which is then parsed and used by the UI.
  
### UI

This is based on the design here: https://docs.google.com/document/d/1hHgEgx1XI9mjha4SArmjtEGwJzQ_XihUPhfhs8mlFhw/edit?tab=t.0#heading=h.pvxbt0ldmbhq

This is the first version of the tabbed App View page - there will be further work to merge the changelog and releases history, based on the users' feedback so far.

[changelog-3.webm](https://github.com/user-attachments/assets/dfa7430c-4e39-4506-833b-42d45d3adcaf)

#### obsolete versions
ignore - just leaving the video if we want to refer back

`https://github.com/user-attachments/assets/f4348a53-b2c4-4a73-a482-1ce688c87566`

`https://github.com/user-attachments/assets/307aa427-8f59-4f81-b0a0-ba0f91ac8a96`


kabaros:
@cooper-joe @amcgee I marked this PR as draft, even though this implementation is ready - I tried to find a decent way to expose the changelog now that we have it, but I am not entirely convinced this is the right direction. I am happy to explore a completely different UX / concept (or few concepts) now that we have the changelog data ready.

cooper-joe:
Thanks for the explanation @kabaros, helpful to get the context. I'm taking a look at this now, exploring some design ideas for viewing and comparing changes.

client/src/components/Versions/VersionsTable/VersionsTable.js
```diff
@@ -40,96 +31,100 @@ const VersionsTable = ({
...
-                    </TableRowHead>
-                </TableHead>
-                <TableBody>
-                    {versions.map((version) => (
-                        <TableRow key={version.id}>
-                            <TableCell>{version.version}</TableCell>
-                            <TableCell className={styles.channelNameCell}>
+        <ol className={styles.versionList}>
+            {versions.map((version) => {
+                const changes = changelogData?.[version.version] ?? []
```
amcgee, line 41: I was a bit surprised to see this as an array... would it not be better to store / display just a markdown text block for each version change as GitHub releases does?

client/src/utils/changelog.js
```diff
@@ -0,0 +1,132 @@
...
+
+            if (line.match(/^\#{3}\s/)) {
+                const logType = this.#getLogType(line)
+
+                lastLogType = logType
+
+                if (logType === 'BREAKING CHANGES') {
+                    const lastVersion = versions[versions.length - 1]
+                    lastVersion.isBreaking = true
+                    versions[versions.length - 1] = lastVersion
```
amcgee, line 76: Why is this necessary?  I think the last line is superfluous, as the object pointed to by that array index is already the `lastVersion` object....

client/src/utils/changelog.js
```diff
@@ -0,0 +1,132 @@
...
+        const lines = this.#rawChangeLog?.split('\n')
+
+        const versions = []
+
+        let lastLogType = ''
+        lines.forEach((line) => {
+            // version header
+            if (line.match(/^\#{1,2}\s/)) {
+                const versionName = this.#getVersion(line)
+                versions.push({ version: versionName, changeSummary: [] })
```
amcgee, line 63: Maybe there should be a `let lastVersion;` outside the `forEach` and this should update it and push into the array?

```suggestion
                lastVersion = { version: this.#getVersion(line), changeSummary: [] }
                versions.push(lastVersion)
```

Then `lastVersion` could just be used by all future iterations until another version header is found, without all the `versions[versions.length - 1]` pointer dereferencing

client/src/utils/changelog.js
```diff
@@ -0,0 +1,132 @@
...
+            }
+
+            // logType: Features, Bug Fixes etc...
+
+            if (line.match(/^\#{3}\s/)) {
+                const logType = this.#getLogType(line)
+
+                lastLogType = logType
```
amcgee, line 71: This could be rewritten as one line

```suggestion
                lastLogType = this.#getLogType(line)
```

amcgee (commented):
Looking good!  Added a few code comments

client/src/components/Versions/VersionsTable/VersionsTable.js
kabaros: it was like that initially to keep the changes in a structured format and display them consistently, to do things like:
- we were displaying the change type in a cell in a table so this was better than doing regex parsing at last minute
- renaming the change types: "feat" to "feature" etc..
- highlighting (or ignoring) translations
- being able to show only bugs or features, or group by change type (never did that but thought it makes sense when we're showing the changes vs instlaled version for example that we'd show all features then all bugs..)

slowly though, the design evolved, and we're not doing any of these .. so I could go back to just having an array of versions, and the changes as a markdown that we display as it is

client/src/api/AppHubAPI.js
```diff
@@ -40,6 +40,10 @@ export default class AppHubAPI {
         { useAuth = false, external = false, params = {} } = {},
         fetchOptions
     ) {
+        if (!path) {
+            return
```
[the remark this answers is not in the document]
kabaros: this is to be able to make this hook conditional, which is inline with how the underlining `useSwr` works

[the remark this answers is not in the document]
kabaros: so that we can conditionally call the /changelog endpoint, only if the app has the flag hasChangelog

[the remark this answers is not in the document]
amcgee: OK got it, sounds good

client/src/components/Versions/VersionsTable/VersionsTable.js
kabaros: @amcgee updated it to just show the raw markdown:

image

PR #748, opened 2025-01-21
feat: implement organisation filter

kabaros:
implements [HUB-161](https://dhis2.atlassian.net/browse/HUB-161)

[organisations-list-with-show-more.webm](https://github.com/user-attachments/assets/9a28d2a2-5ded-45da-a844-c9b27c0eef75)

### DIFF
--- a/server/src/routes/v2/organisations.js
+++ b/server/src/routes/v2/organisations.js
@@ -71,6 +71,24 @@ module.exports = [
     },
     {
         method: 'GET',
+        path: '/v2/organisations-with-apps',
+        config: {
+            auth: false,
+            tags: ['api', 'v2'],
+            cache: {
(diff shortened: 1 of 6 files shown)


kabaros:
ps: PR against changelog branch for now

server/src/routes/v2/organisations.js
```diff
@@ -69,6 +69,24 @@ module.exports = [
             return orgs
         },
     },
+    {
+        method: 'GET',
+        path: '/v2/organisations-with-apps',
```
kabaros, line 74: not sure about the route name - the API is not RESTful anyhow, so went for simplicity

Birkbjo: I would expect this to be under `/organisations`. And maybe a resource under that?

PR #749, opened 2025-02-06
chore: expose appview without the shell to embed in app-management app

kabaros:


client/src/AppHub.js
```diff
@@ -32,34 +33,54 @@ const AppHub = () => (
...
-                                    exact
-                                    path="/verify/org"
-                                    component={OrganisationInvitation}
-                                />
-                                {/* No-match route - redirect to index */}
-                                <Route render={() => <Redirect to="/" />} />
-                            </Switch>
-                        </main>
+                        <Switch>
+                            <Route
```
kabaros, line 37: this is a route that doesn't have the shell (the headre and sidebar..) of AppHub, that can be integrated as an iframe in app-management

